PR #9, opened 2020-12-08
Option to include source code to compiler artifact

blitz-1306:
## Changes
- [x] Added CLI option `--with-sources` to include source files content to the compiler artifact.
- [x] Added an integration test for the feature.

## Notes
The option requires `--raw` to be specified. Usage:
```console
sol-ast-compile sample.sol --raw --with-sources
```

Regards.


cd1m0 (approved):
Awesome!

PR #10, opened 2020-12-10
Fix missing anonymous keyword when writing event definitions.

cd1m0:
What it says in the title :).


blitz-1306 (approved):
Look good. Thanks for the fix.

PR #15, opened 2021-01-15
Sanitize referenced declarations for builtin references

blitz-1306:
## Preface
The compiler sometimes provides invalid positive ids for some builtin references. For example in `abi.encode()` the identifier `abi` gets an id that +1 higher than the highest id in the current context.

This becomes a problem if we try to make a new node in the current AST context. Then all of a sudden the `abi` identifier stops being a builtin and now refers to something real.

One way to deal with this, is to do another pass after finalizing (or during finalizing), in which to set all invalid `referencedDeclaration`s to `-1`.

## Changes
- Introduced `BuiltinReferencedDeclarationNormalizer` to tweak `referencedDeclaration` property values, that refer to an inexisting ids. They would have `-1` as a value.
- Added it to postprocessor pass for `Identifier`, `MemberAccess`, `IdentifierPath` and `UserDefinedTypeName` nodes.

## Concerns
This is a bit destructive, though… If compiler will ever do some changes, we would have to adjust logic to only do such cleanup for certain compiler versions. But that concern is not yet have any real use case. 

Regards.

### DIFF
--- a/src/ast/postprocessing/builtin_referenced_declaration_normalizer.ts
+++ b/src/ast/postprocessing/builtin_referenced_declaration_normalizer.ts
@@ -1,0 +1,32 @@
+import { ASTNode } from "../ast_node";
+import { ASTContext, ASTNodePostprocessor } from "../ast_reader";
+import { Identifier } from "../implementation/expression/identifier";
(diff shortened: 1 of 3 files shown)


cd1m0 (approved):
LGTM

PR #16, opened 2021-01-28
Provide source maps for written (sourced) ASTNodes

blitz-1306:
## Changes
- Tweaks to AST writing components to provide a way to obtain a source map for the written nodes:
  - [X] [breaking] The `ASTNodeWriter` is now an abstract class instead of interface. It introduces `writeInner()` and `writeWhole()` methods, that are regulated by descendant classes and provides more control over the node writing.
  - [X] The `ASTNodeWriter` descendant classes are more accurately check the necessity if additional parentheses are needed to maintain the evaluation order of expressions. The overall amount of emitted "failsafe" parentheses was reduced.
  - [X] The `ASTWriter` also has few new methods: `desc()` for getting intermediate representation of sources and `descToSourceString()` to convert mentioned representation to a final source string. The method `write()` received new optional argument `sourceMap`, that allowsto pass the map object to store `ASTNode -> [offset, length]` source mapping coordinates in the written final source.
- Tweaks for the `ASTNode` descendant classes:
  - [X] Tweaked order of `FunctionDefinition` children: the return `ParameterList` now will appear after `OverrideSpecifier` and `ModifierInvocation[]`. It is more accurately follow the location of these parts in the source code: modifiers and `override`s are preceeding the `returns(...)` part.
  - [X] [breaking] The `ForStatement.vInitializationExpression`'s type now allowed to contain `ExpressionStatement`. Previously we had an oversight, that was giving a false assumtion that only `VariableDeclarationStatement`s are allowed there. This is now fixed.
- Changes in tests:
  - [X] The written source are now tested by an integration test with compiler invocation. This means that we are sure, that we write code, that is able to be compiled.
  - [X] The generated source maps are verified by an integration test with few snapshots.
  - [X] The generated source maps are verified by an integration test to match  the SolcJS source maps (except the bugs and edge cases of the Solc itself).
  - [X] Tweaked affected snapshots of other tests to match new expectations for a written sources.

Regards.


src/ast/writing/writer.ts
```diff
@@ -63,3 +73,74 @@ export class ASTWriter {
...
+
+        const sibling = node.previousSibling;
+
+        if (sibling) {
+            const [startS, lenS] = this.getPrecomputedCoords(sibling, computed);
+
+            offset = startS + lenS - startP;
+        }
+
+        const index = sourceP.indexOf(sourceN, offset);
```
cd1m0, line 132: Why is this neccessary? Shouldn't it always be that `index == offset`?

blitz-1306: The `index` of the subnode is not the same as `offset` to start search. We also need to skip the whitespace and comments. Lets consider an example:

```solidity
contract A {
    function some() pure public {
        uint a = 1;
        // some comment
        uint b = 2;
    }
}
```

To pick location of `uint b = 2;`, we need to seach in source of the whole function definition. However, we need to skip the indentation space and comment here. The `offset` would replicate that we start search in function definition source just after the declaration `uint a = 1;` to skip collisions.

blitz-1306: Also added comments with clarification in https://github.com/ConsenSys/solc-typed-ast/pull/16/commits/0e89a4dc7faef6a2a63cb57d7568428e7db4d388.

cd1m0 (approved):
LGTM

PR #17, opened 2021-02-01
Some nits and fixes to help with the source map pr

cd1m0:
This PR does several things:

 - Fix source mapping bug due to bad ordering of children of FunctionDeinition (OverrideSpecifier was out of order).
 - Fix source mapping off-by-one bug in ForStatement due to removal of semicolon;
 - Fix writing logic to not emit extra parenthesis around top-level expressions (allows for reproducible repeated writes); 
 - Add WIP test for source mappings
 
 There is more work to make that test run cleanly, but I think its useful as it already helped me find 2 bugs. To get it fully working we need to:
 
  - fix the source-map of any node that has a StructuredDocumentation child to NOT include that child in its source map
  - fix remaining off-by-one issues


src/ast/writing/ast_mapping.ts
```diff
@@ -191,333 +221,393 @@ class LiteralWriter implements ASTNodeWriter {
...
+class ForStatementWriter extends CompoundStatementWriter<ForStatement> {
+    writeInt(node: ForStatement, writer: ASTWriter): SrcDesc {
+        return writer.desc(
+            "for (",
+            ...(node.vInitializationExpression === undefined
+                ? ["; "]
+                : [node.vInitializationExpression, " "]),
+            node.vCondition,
+            "; ",
+            node.vLoopExpression,
```
blitz-1306, line 550: I believer these two are also [optional](https://github.com/ConsenSys/solc-typed-ast/blob/master/test/samples/solidity/statements/for_050.sourced.sol#L82).

src/ast/writing/writer.ts
```diff
@@ -1,14 +1,61 @@
...
+    /**
+     * Add the node to the descrioption generated by `writeInt`, any additional 'wrappings' and return it. Usually `write` is responsible
+     * for adding semicolons and documentation, since these are not generally part of source mappings.
+     *
+     * For example given this source:
+     * `a = 1;`
+     *
+     * `ExpressionStatement.writeInt` would return the following desc:
+     *
+     * [[Assignment#3, [[Identifier#1, ["a"]], " = ", [Literal#2, ["1"]] ]]]
```
blitz-1306, line 47: The most concerning part here is the amount of object allocations for the deeply nested trees. Unsure how much memory it will consume.

test/integration/sol-ast-compile/writing_compiles.spec.ts
```diff
@@ -0,0 +1,79 @@
...
+        it("Exit code is valid", () => {
+            expect(exitCode).toEqual(0);
+        });
+
+        it("STDERR is empty", () => {
+            expect(errData).toContain("");
+        });
+
+        it("STDOUT produces correct bytecode", () => {
+            fse.writeFileSync("tmp.sol", outData, { encoding: "utf8" });
```
blitz-1306, line 67: This can be done in async fashion, so we better to not lock tests. Also, I guess we can jus compile from STDIN instead with using `--raw`. Would be cool.

blitz-1306 (commented):
Overall, I like the flexibility. However, there are few points that we better discuss first:
- Memory usage and amount of the object allocations.
- I guess API interface may be cleaner.
- The bytecode acceptance test (I like the idea, but we can go with a bit different implementation).

Lets think about it a bit more and discuss properly.

src/ast/writing/ast_mapping.ts
cd1m0: The `writer.desc()` function supports optional arguments (i.e. undefined and null)

cd1m0: This one got a special case just cause I wanted to write "; " instead of "" when its omitted.

src/ast/writing/writer.ts
cd1m0: Its true that compared to the other scheme we are allocating a lot of small arrays here. On the other hand we are omitting the string scanning that was done in ASTSourceMapComputer before, and getting more precise and fine-grained results. So we are trading memory for speed and precision. Perhaps we need to add some memory profiling to track how much each stage consumes?

test/integration/sol-ast-compile/writing_compiles.spec.ts
cd1m0: Yeah, that test is kinda hacky, and I added it just to help my development. If you don't like it, we can remove it, or feel free to do any changes to it? Originally I tried extracting the bytecodes and comparing them. However as it turned out they weren't exactly the same. I didn't have time to debug and understand whats wrong, so I just changed the test to just make sure it compiles. The string is out-of-date - will fix

cd1m0: I don't know how to pipe input into a node.js child process to get it working with `--stdin`. Could you fix it @blitz-1306 ?

cd1m0:
I got the PR to the point where the only cases where we differ from the solc ast maps are:

 - when there is StructuredDocumentation. There is inconsistent handling of comments in the solc ast between `/*** ***/` and `///`. This offsets the written code from the original code by several characters. Note that the source fragments still align - they are just 1-2 characters off
 - `payable` typenames. Solc source maps include a parenthesis, ours don't. Likely a bug in solc
 - function typename variable defs. Solc's source maps **incorrectly** include the variable name in the source map of the type name
 - some nodes that don't have any corresponding text are not even added to our source map.
 
Given that most of the cases where we don't align are related to solc bugs, I'm pretty happy with this now.

PR #19, opened 2021-02-24
Bug in the writing out of import directives

cd1m0:
Small bugs in writing out import directives:

   - when writing out `x as y` directives, we were incorrectly writing
     out the original definition instead of the name given
   - when the `foreign` field was an Identifier, we were forgetting to
     add it to the source map

### DIFF
--- a/src/ast/writing/ast_mapping.ts
+++ b/src/ast/writing/ast_mapping.ts
@@ -1131,21 +1131,32 @@ class ContractDefinitionWriter extends DocumentedNodeWriter<ContractDefinition> 
 }
 
 class ImportDirectiveWriter extends ASTNodeWriter {
-    writeInner(node: ImportDirective): SrcDesc {
+    writeInner(node: ImportDirective, writer: ASTWriter): SrcDesc {
         if (node.unitAlias) {
             return [`import "${node.file}" as ${node.unitAlias};`];
         }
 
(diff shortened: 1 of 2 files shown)


src/ast/writing/ast_mapping.ts
```diff
@@ -1131,21 +1131,32 @@ class ContractDefinitionWriter extends DocumentedNodeWriter<ContractDefinition>
...
+                    entries.push(desc);
+                } else {
+                    const symbol =
+                        origin instanceof ImportDirective ? origin.unitAlias : origin.name;
+                    entries.push([alias !== undefined ? symbol + " as " + alias : symbol]);
+                }
             }
 
-            return [`import { ${entries.join(", ")} } from "${node.file}";`];
+            return [`import { `, ...flatJoin(entries, ", "), `} from "${node.file}";`];
```
blitz-1306, line 1159: Missing space before `} from "${node.file}";`. It should have preceeding space char to comply with previous behavior.

blitz-1306 (changes requested):

src/ast/writing/ast_mapping.ts
cd1m0: Fixed in e481519

blitz-1306 (approved):
LGTM

PR #20, opened 2021-02-24
Fix for implicit parentheses

cd1m0:
Sometimes when we write-out a UnaryOperation/BinaryOperation/Conditional we add implicit parenthesis. So `BinaryOperation(Identifier('x'), Identifier('y'))` may be written out as `(x + y)`. Currently the source map we compute includes the parentheses. However the bytecode map for this same expression won't include the parentheses. So this PR changes the behavior of the computed source map to not include implicit parentheses.


blitz-1306 (approved):
LGTM

PR #22, opened 2021-03-02
Support for Solidity 0.8.1 release

blitz-1306:
## Premise
Support for the Solidity 0.8.1 release is somehow went more complex than expected. There are two issues https://github.com/ethereum/solc-js/issues/513 and https://github.com/ethereum/solidity/issues/10868 in upstream dependencies that forced us to reconsider current approach to handle different versions of the compiler. As NPM packages of SolcJS are sometimes mispublished and this is not fixable of behalf of Solidity team, we would switch to loading of original compiler snapshots from https://binaries.soliditylang.org/. That would allow us to support legacy compiler versions and have proper recent ones, that are fixable by the Solidity team (as NPM do not allow to update published packages). This complication forces us to remove a set of dependencies and introduce compiler snapshots downloading and use during the common execution flow. Even though there is no major API changes, we did changed a lot of things in compiling logic to avoid further issues. This may lead to unexpected differences in behavior.

## Changes
- [X] Removed `solc-<version>` dependencies. Left only latest one to be able to load compiler snapshot. Previous approach is no longer reliable.
  - Added `setup/download_compilers.js` script to download compilers to local directory. Added npm script command `download-compilers` and `npm run download-compilers` build step to download compilers during the package installation process.
  - Tweaked compile logic to load and use compiler snapshots instead.
- [X] Added `Throw` node reading rule in modern (compact) AST configuration, as this node type appears in modern (compact) AST in legacy compiler snapshots.
- [X] Updated dependencies.
- [X] Added support for Solidity 0.8.1 release.
- [X] Tweaked `TryCatchClauseWriter` to not emit double spaces if `catch` was provided without parameters.
- [X] Tweaked affected tests.

## Notes
- Such changes may lead to unexpected differences in behavior. Also we have a lot of changes in dependencies. We better bump major version for a next package release.
- Unfortunately, such chnages also leading to a performance degradation (due to compiler side-loading the compiling is ~50% slower).

## Related links
- [Solidity 0.8.1 release notes](https://github.com/ethereum/solidity/releases/tag/v0.8.1)
- [Solidity 0.8.1 release blog post](https://blog.soliditylang.org/2021/01/27/solidity-0.8.1-release-announcement/)

Regards.


src/compile/utils.ts
```diff
@@ -236,24 +235,10 @@ export function compile(
...
 
-    if (satisfies(version, "0.4")) {
-        const input = createCompiler04Input(fileName, content, remapping);
-        const output = compiler.compile(input, 1, finder);
-
-        return output;
-    }
-
-    if (satisfies(version, "0.5")) {
-        const input = createCompiler05Input(fileName, content, remapping);
```
cd1m0, line 249: Why did we have separate `createCompiler05Input` and `createCompiler04Input` functions in the past?

test/samples/solidity/latest_08.sol
```diff
@@ -35,3 +35,23 @@ contract UsesNewAddressMembers {
...
+
+contract CatchPanic {
+    function test() public {
+        UsesNewAddressMembers c = new UsesNewAddressMembers();
+
+        try c.test() {
+            
+        } catch Error(string memory reason) {
+            revert(reason);
+        } catch Panic(uint _code) {
```
cd1m0, line 47: Where is `Panic` defined? I thought in 0.8.1 you had to define the error types?

cd1m0 (approved):
Looks good! I just have some questions cause I don't understand everything.

src/compile/utils.ts
blitz-1306: For legacy package versions the `sources` property had following format:
```
```

test/samples/solidity/latest_08.sol
blitz-1306: `Panic` is indicated as velur of `errorName` property of `TryCatchClause` node: https://github.com/ConsenSys/solc-typed-ast/blob/master/src/ast/implementation/statement/try_catch_clause.ts#L10

blitz-1306:
After a few round of discussion we decide to put this on hold, mostly beause of degraded performance of compiler invocation. Tests are taking much more time, than wit previous approach.

I will close this PR, but will leave a branch for some time (just in case).

PR #23, opened 2021-03-18
Support field names specification when writing source for the FunctionCall nodes

blitz-1306:
## Changes
- [x] Fix non-covered case of writing the function calls with custom order of the args.
- [x] Tweaked writer edge case sourcing samples.

Regards.


cd1m0 (approved):
LGTM

PR #25, opened 2021-03-29
Support for Solidity 08.1, 0.8.2 and 0.8.3

blitz-1306:
## Changes
- [x] Bumped development dependency versions: `semver` major dependency and the development dependencies.
- [x] Added support for **Solidity 0.8.2** and **Solidity 0.8.3** releases on compile level.
- [x] Added support for **Solidity 0.8.2** and **Solidity 0.8.3** releases on AST level:
  - Introduced `documentation` for `Statement` descendants.
  - Adopted **Solidity 0.8.1**-related changes:
    - Codebase from #22 as they are part of the **Solidity 0.8.2** and **Solidity 0.8.3** releases.
    - Custom package (https://www.npmjs.com/package/solc-0.8.1-fixed) from #27.
  - Introduced `nameLocation` for definition/declaration nodes. Introduced `parseSourceLocation()` function and `SourceLocation` type for easy parsing of `nameLocation`s.
  - Introduced `ModifierInvocationKind` enum and `kind` property for `ModifierInvocation` node.
- [x] Tweaked AST-to-source writing components to respect the changes.
- [x] Tweaked `ASTNodeFactory` to respect the changes.
- [x] Tweaked related test suites and samples.

## Notes
- This PR causes backward incompatibility due to changes in node constrcutor signatures and arguments for `ASTNodeFactory` methods.
- The legacy AST reading is adusted just to not crash. **It is only in maintenance mode until further notifications.** New functionality is `undefined` by default of legacy AST nodes and related processing logic.
- `StructuredDocumentation` reconstruction is not enabled for `Statement` descendant nodes. So documentation is `string` only.

## Related links
- [Solidity 0.8.1 release notes](https://github.com/ethereum/solidity/releases/tag/v0.8.1)
- [Solidity 0.8.1 release blog post](https://blog.soliditylang.org/2021/01/27/solidity-0.8.1-release-announcement/)
- [Solidity 0.8.2 release notes](https://github.com/ethereum/solidity/releases/tag/v0.8.2)
- [Solidity 0.8.2 release blog post](https://blog.soliditylang.org/2021/03/02/solidity-0.8.2-release-announcement/)
- [Solidity 0.8.3 release notes](https://github.com/ethereum/solidity/releases/tag/v0.8.3)
- [Solidity 0.8.3 release blog post](https://blog.soliditylang.org/2021/03/23/solidity-0.8.3-release-announcement/)
- https://github.com/ethereum/solidity/issues/11194
[...]

PR #27, opened 2021-04-09
Support Solidity 0.8.1

cd1m0:
Support for 0.8.1 using custom built npmjs package (https://www.npmjs.com/package/solc-0.8.1-fixed). This is mostly a copy of https://github.com/ConsenSys/solc-typed-ast/pull/22 except for the use of the new package instead of downloading solc binaries.


blitz-1306:
While it looks good to merge, let's close it in favor of #25. As 0.8.1-related changes are already there, we can add a single line to `package.json`:
```
"solc-0.8.1": "npm:solc-0.8.1-fixed@0.8.1",
```
So, the most valuable part here is the published package. Thank you for providing it.

PR #29, opened 2021-04-15
Migrate some features from Scribble to solc-typed-ast

cd1m0:
This PR adds the following helper functions:
- [x] `replaceNode(oldNode, newNode)` replaces a given node in the tree with another
- [x] `isSane(units, ctx)` checks whether a given AST tree is "sane". This involves making sure that all reachable nodes are part of the same context, that parent/child relationships are correctly set, and certain property pairs are related. This is useful to check that a re-written, or manually constructed AST tree looks correct

We also add a new integration test for `replaceNode()` that checks that we can replace most nodes in the samples we have, and produce a sane tree.

### DIFF
--- a/src/ast/sanity.ts
+++ b/src/ast/sanity.ts
@@ -218,8 +218,7 @@ function checkFieldAndVFieldMatch<T extends ASTNode, K1 extends keyof T, K2 exte
  */
 function checkDirectChildren<T extends ASTNode>(node: T, ...fields: Array<keyof T>): void {
     const directChildren = new Set(node.children);
-
-    const computedChildren: Set<ASTNode> = new Set();
+    const computedChildren = new Set<ASTNode>();
 
     for (const field of fields) {
         const val = node[field];
@@ -236,6 +235,7 @@ function checkDirectChildren<T extends ASTNode>(node: T, ...fields: Array<keyof 
                     )} child of ${pp(val.parent)}`
                 );
             }
+
             computedChildren.add(val);
         } else if (val instanceof Array) {
             for (let i = 0; i < val.length; i++) {
@@ -249,7 +249,7 @@ function checkDirectChildren<T extends ASTNode>(node: T, ...fields: Array<keyof 
                     throw new Error(
                         `Field ${field} of ${pp(
                             node
-                        )} is neither an ASTNode nor an array of ASTNode or nulls - instead array containing ${el}`
+                        )} is expected to be ASTNode, array or map with ASTNodes - instead array containing ${el}`
                     );
                 }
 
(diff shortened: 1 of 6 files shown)


cd1m0:
It seems that my line coverage is unsuffucient - 94.9% for a minimum of 95%. Also it dropped from 96% last time. @blitz-1306  is there some place where I can find the line-by-line coverage?

blitz-1306:
The `.npmrc` file is blocking upload, by exiting with error before upload begins. We can reconfigure it in such way that codecov would result an error instead of nyc/mocha, so we still would get report. My bet would be that `isSane()` / `replaceNode()` are never invoked in the tests, or they are not sufficiently covered.

You can also look though the local coverage report by running tests and then browsing `<solc-typed-ast repo path>/coverage/index.html`.

src/ast/sanity.ts
```diff
@@ -0,0 +1,623 @@
...
+ */
+export function isSane(unit: SourceUnit, ctx: ASTContext): boolean {
+    try {
+        checkSane(unit, ctx);
+    } catch (e) {
+        if (e instanceof InsaneASTError) {
+            console.error(e);
+
+            return false;
+        }
```
blitz-1306, line 617: Do we intentionally silent the `InsaneASTError`? Or is it for debugging purposes?

blitz-1306 (commented):
Thanks for the contribution. A minor question.

src/ast/sanity.ts
cd1m0: Yeah - that error is just for debugging purposes - to communicate to `isSane` (or any other consumer) what exactly was the problem. If you want to do something else other than silence then you call `checkSane` directly

blitz-1306 (dismissed):

blitz-1306:
@cd1m0 I did some tweaks to the  code:
- `replaceNode()` and `checkDirectChildren()` can now process properties of simple `Map` type, that have `ASTNodes` as values. This covers situation with `FunctionCallOptions` and some other nodes, however we still have and edge-case with `ImportDirective`.
- `checkSane()` now properly processed nodes with `documentation: string | StructuredDocumentation | undefined`.

Coverage is now on acceptable level, however we can do a bit more passes over the codebase to clean it up.

blitz-1306:
@cd1m0 I removed the part of the `replaceNode()`, that was responsible for getter/setter lookup in prototype chain. It is never hit due to the only candidate at the moment is `ContractDefinition` with `documentation` setter. However, even this setter is not hit as the field `ownChildren` is handled by `Object.getOwnPropertyDescriptors()`. So children are patched without any necessity to interact with setters. This also bumbed coverage a bit.

blitz-1306 (approved):
Looks good to merge now.

PR #33, opened 2021-04-27
Support for Solidity 0.8.4

blitz-1306:
## Changes
- [x] Bumped dependency versions: `web3-eth-abi` to `^1.3.5`, also development dependencies to use latest tools.
- [x] Added support for **Solidity 0.8.4** release on compile level (`solc@0.8.4`).
- Added support for **Solidity 0.8.4** release on AST level:
  - [x] Introduced new statement node `RevertStatement`.
  - [x] Introduced new definition node `ErrorDefinition`.
  - Tweaked `ContractDefinition`:
    - Introduced `usedErrors` and `vUsedErrors` properties to track used errors.
    - Introduced `vErrors` property.
    - Tweaked constructor to accept additional required array argument for `usedErrors` (**a breaking change**).
  - Tweaked `SourceUnit`:
    - Tweaked `ExportedSymbol` type to also allow `ErrorDefinition`s (this also will affect imports, **a breaking change**).
    - Introduced `vErrors` property.
- [x] Updated `NODE_LIST` to mention new nodes.
- [x] Tweaked sanity checking logic to respect the changes.
- [x] Tweaked AST-to-source writing components to respect the changes (see **Concerns** section).
- [x] Tweaked `ASTNodeFactory` to respect the changes.
- [x] Tweaked related test suites and samples.

## Concerns
I believe that `YulHexStringLiteral`s are undeкworked in compiler output:
- I did not found a convenient way to distiguish common Yul string literal from `YulHexStringLiteral`. Due to this, it is really hard to make a decision how to write any string literals (common string or hex). The Solidity AST have `Literal.kind`, that allows to determine how to write node properly, but Yul does not.
- I did not found a proper solution to analyze `YulHexStringLiteral` content values to write them as `hex""` and get same output, as input was. Maybe there is something happening during JSON serialization/deserialization, that causes content fragmentation. The [most stable approach](https://stackoverflow.com/a/57391629) I recall is still produces something, that is different from expectations.

I will create related issue()s on Solidity repository to figure out how to solve this. The [latest_08.sourced.sol](https://github.com/ConsenSys/solc-typed-ast/blob/1f43580953b0b83308facc5cb13328fa009b95ab/test/samples/solidity/latest_08.sourced.sol) is broken for now.

## Notes
- This PR causes backward incompatibility due to changes in node constrcutor signatures and arguments for `ASTNodeFactory` methods.
- The legacy AST reading is adusted just to not crash. **It is only in maintenance mode until further notifications.** New functionality is `undefined` by default of legacy AST nodes and related processing logic.

## Related links
- [Solidity 0.8.4 release notes](https://github.com/ethereum/solidity/releases/tag/v0.8.4)
- [Solidity 0.8.4 release blog post](https://blog.soliditylang.org/2021/04/21/solidity-0.8.4-release-announcement/)
- [Solidity blog article on custom errors](https://blog.soliditylang.org/2021/04/21/custom-errors/)
- [Git commit of `YulHexStringLiteral`s support](https://github.com/ethereum/solidity/commit/7eb5e27e54c3ffc0e0a1e6d9b6c50e992df6886b)

Regards.


src/ast/implementation/declaration/error_definition.ts
```diff
@@ -0,0 +1,61 @@
...
+
+    /**
+     * The source range for name string
+     */
+    nameLocation?: string;
+
+    /**
+     * Optional documentation appearing above the error definition:
+     * - Is `undefined` when not specified.
+     * - Is type of `string` when specified and compiler version is older than `0.6.3`.
```
cd1m0, line 20: There are no errors on compilers older tham 0.8.4 so should we allow `string` here?

src/ast/implementation/declaration/error_definition.ts
```diff
@@ -0,0 +1,61 @@
...
+    }
+
+    get children(): readonly ASTNode[] {
+        return this.pickNodes(this.documentation, this.vParameters);
+    }
+
+    /**
+     * Reference to its scoped contract
+     */
+    get vScope(): ContractDefinition {
```
cd1m0, line 58: Looking at `SourceUnit` it looks like you can have `ErrorDefinition`s at the SourceLevel too

src/ast/implementation/statement/revert_statement.ts
```diff
@@ -0,0 +1,29 @@
+import { ASTNode } from "../../ast_node";
+import { FunctionCall } from "../expression/function_call";
+import { Statement } from "./statement";
+
+export class RevertStatement extends Statement {
+    /**
```
cd1m0, line 6: In 0.8.4 would `revert()` appear as the new `RevertStatement` node, or as an old-style `ExpressionStatement(FunctionCall...)`?

src/ast/modern/error_definition_processor.ts
```diff
@@ -0,0 +1,31 @@
...
+        const [id, src, type] = super.process(reader, config, raw);
+
+        const name: string = raw.name;
+        const nameLocation: string | undefined = raw.nameLocation;
+
+        let documentation: string | StructuredDocumentation | undefined;
+
+        if (raw.documentation) {
+            documentation =
+                typeof raw.documentation === "string"
```
cd1m0, line 22: Given that this should happen in only 0.8.4 and newer, shouldn't this always be a StructuredDocumentation?

cd1m0 (changes requested):
Couple of nits, otherwise LGTM.

src/ast/implementation/statement/revert_statement.ts
blitz-1306: As `RevertStatement`. It is a new standalone expression node like `EmitStatement`.

src/ast/implementation/declaration/error_definition.ts
blitz-1306: This is for compatibility reasons with legacy behavior. Sometimes it is just _convenient_ to pass string, especially if you do a lot of node creation.

I will fix the comment here, stating that is can be a string for compatibility reasons.

src/ast/modern/error_definition_processor.ts
blitz-1306: See previous [related comment](https://github.com/ConsenSys/solc-typed-ast/pull/33#discussion_r623759395).

src/ast/implementation/declaration/error_definition.ts
blitz-1306: Good catch! Fixed in https://github.com/ConsenSys/solc-typed-ast/pull/33/commits/04b57a405cf1aafedf44c93a7fef2e9bcb064287.

cd1m0 (approved):

PR #51, opened 2021-07-15
Fix canonical signature hashes (`selector`s) computation

### LINKED ISSUE #49: Incorrect computation of canonical signature hashes (`selector`s) of `FunctionDefinition`
## Environment
Any

## Scenario
- Create file `sample.sol` with following content:
  ```solidity
  library Some {
      struct Item {
          uint id;
      }

      struct Abc {
          Item item;
      }

      struct Def {
          Item item;
      }

      function id(Abc memory data) pure public returns(uint) {
          return data.item.id;
      }

      function id(Def memory data) pure public returns(uint) {
          return data.item.id;
      }
  }

  contract Test {
      using Some for Some.Abc;
      using Some for Some.Def;

      function verify() pure public {
          Some.Abc memory a = Some.Abc({ item: Some.Item(1) });
          Some.Def memory b = Some.Def({ item: Some.Item(2) });

          assert(a.id() == 1);
          assert(b.id() == 2);
      }
  }
  ```
- Run `solc-ast-compile` on `sample.sol`:
  ```bash
  sol-ast-compile sample.sol --xpath '//FunctionDefinition' --depth 0
  ```

## Expected behavior
Expected following `canonicalSignatureHash`es:
- For `id(Abc memory data)` - `7c503cd6`
- For `id(Def memory data)` - `9c9d982f`

## Current behavior
Got following `canonicalSignatureHash`es:
- For `id(Abc memory data)` - `fe412341`
- For `id(Def memory data)` - `fe412341`

## Description
This is due to legacy behavior of `VariableDeclaration`s `canonicalSignatureType` getter: https://github.com/ConsenSys/solc-typed-ast/blob/9eb5c6ee3886788bc9ad4045655719c72acedbfd/src/ast/implementation/declaration/variable_declaration.ts#L163-L169 It produces following signatures to hash for both of the functions - `id(((uint256)))`. This results same signature hash value for different function overloads.

While it may be valid for legacy Solidity, it also is invalid for ABI encoder v2, that is default for recent Solidity versions.

Regards.

blitz-1306:
## Tasks
Fixes #49.

## Changes
- [x] Fixed canonical signature hash computation. 
- [x] Fixed affected test samples. Added related test sample.

## Notes
This fix may look silly, however there is a reason to keep things "as simple as possible". It might be a bit surprising, that compiler catches and restricts possible signature clashes for a following case:
- **a.sol**
  ```solidity
  library A {
      struct S {
          uint value;
      }
  }
  ```
- **b.sol**
  ```solidity
  import { A as T } from './a.sol';

  library A {
      struct S {
          uint value;
      }

      function get(S memory s) pure public returns (uint) {
          return s.value;
      }

      function get(T.S memory s) pure public returns (uint) {
          return s.value;
      }
  }
  ```
Compiler produces following:
```text
TypeError: Function overload clash during conversion to external types for arguments.
 --> test.sol:12:5:
  | 12
  | function get(T.S memory s) pure public returns (uint) {
  | ^ (Relevant source part starts here and spans across multiple lines).
```
This proves that we should not encounter cases when symbol aliasing is causing additional complications.

Regards.

### DIFF
--- a/src/ast/implementation/declaration/variable_declaration.ts
+++ b/src/ast/implementation/declaration/variable_declaration.ts
@@ -161,11 +161,11 @@ export class VariableDeclaration extends ASTNode {
             const declaration = type.vReferencedDeclaration;
 
             if (declaration instanceof StructDefinition) {
-                const signatures = declaration.vMembers.map(
-                    (member) => member.canonicalSignatureType
-                );
+                if (declaration.vScope instanceof ContractDefinition) {
+                    return declaration.vScope.name + "." + declaration.name;
+                }
 
-                return "(" + signatures.join(",") + ")";
+                return declaration.name;
             }
 
             if (declaration instanceof ContractDefinition) {
--- a/src/ast/implementation/declaration/variable_declaration.ts
+++ b/src/ast/implementation/declaration/variable_declaration.ts
@@ -1,5 +1,5 @@
 import { ASTNode } from "../../ast_node";
-import { DataLocation, Mutability, StateVariableVisibility } from "../../constants";
+import { ContractKind, DataLocation, Mutability, StateVariableVisibility } from "../../constants";
 import { encodeSignature } from "../../utils";
 import { Expression } from "../expression/expression";
 import { OverrideSpecifier } from "../meta/override_specifier";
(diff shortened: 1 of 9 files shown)


cd1m0:
After some discussion with Pavel we found that if we modify the `canonicalSignature` and `canonicalSignatureHash` getters to always return the V2 encoding of struct types, then we would break other tools that rely on the AST for this sample:

```pragma solidity 0.6.12;
pragma experimental ABIEncoderV2;

contract Foo {
    struct S {
        uint x;
    }
   
    function foo(S calldata s) external {
    }
    
    function main() public {
        assert(Foo.foo.selector == bytes4(keccak256("foo((uint256))")));
    }
}
```

Fundamentally, the problem is that the exact value of the `canonicalSignature` field for any AST depends on 2 things:

1. The exact version of the compiler used to compile the ast
2. For versions older than when ABIEncoderV2 became the default, on the presence of `pragma experimental ABIEncoderV2`.

So the responsibility falls on the caller, to specify what version of the ABI they want the `canonicalSignature` for. So perhaps we can do something like

1. Introduce a new `ABIEncoderVersion` enum, that currently just has versions 1 and 2
2. Turn the `canonicalSignature`, `canonicalSignatureHash` and `interfaceId` getters into methods that take an `ABIEncoderVersion` argument and return the proper argument (or potentially undefined if for example V1 encoder doesn't support some function type supported by V2)

This way its the responsibility of the caller, depending on the exact version and the presence of experimental pragmas to decide.

blitz-1306:
@cd1m0 See https://github.com/ConsenSys/solc-typed-ast/issues/49#issuecomment-882316362. I implemented fix differently and I guess it is no longer "breaking". I just originally missed the difference in logic. Asking for a re-review.

src/ast/implementation/declaration/variable_declaration.ts
```diff
@@ -158,33 +158,60 @@ export class VariableDeclaration extends ASTNode {
...
 
-                for (let n = 8; n <= 32; n += 8) {
-                    if (length < 2 ** n) {
-                        return "uint" + n;
-                    }
+                if (
+                    declaration instanceof StructDefinition ||
+                    declaration instanceof EnumDefinition
+                ) {
+                    return declaration.vScope instanceof ContractDefinition
```
cd1m0, line 180: You can use `getUserDefinedTypeFQName` from `utils.ts` here.

src/ast/implementation/declaration/variable_declaration.ts
```diff
@@ -158,33 +158,60 @@ export class VariableDeclaration extends ASTNode {
...
+                }
+
+                if (declaration instanceof ContractDefinition) {
+                    return "address";
+                }
+
+                if (declaration instanceof EnumDefinition) {
+                    const length = declaration.children.length;
+
+                    for (let n = 8; n <= 32; n += 8) {
```
cd1m0, line 200: Is there a language limit for enums to not have more than 2**32 entries? Otherwise this should be `n<=256` no? (It doesn't really matter much - no one will declare an enum with even close to 2**32. Just asking

cd1m0 (changes requested):
Awesome work! Just one nit and one q

src/ast/implementation/declaration/variable_declaration.ts
blitz-1306: See https://github.com/ethereum/solidity/releases/tag/v0.8.0:
> Type System: Disallow enums with more than 256 members.

We use common JS integer here: it would loose precision for too large numbers. We can switch to `BigInt` here, however, I do not see the reason. We never encountered this issue (as you also spotted).

src/ast/implementation/declaration/variable_declaration.ts
blitz-1306: Will update.

blitz-1306: Fixed in https://github.com/ConsenSys/solc-typed-ast/pull/51/commits/08fe8e8b015fcfea89c825ef869fc18a7aa92d58

cd1m0 (approved):

PR #63, opened 2021-09-16
Adopt internal types from Scribble

blitz-1306:
## Premise
This PR migrates and adopts [Scribble](https://github.com/ConsenSys/scribble) functionality, related to builtin and internal types. After discussion with maintainer, we decided that it would be reusable by other consumers. Related items are following:
- [`getTypeForCompilerVersion()`](https://github.com/ConsenSys/scribble/blob/d4da2edac356793067afb36dd2588d4f91746a84/src/spec-lang/tc/typecheck.ts#L695-L706)
- [`BuiltinStructType`](https://github.com/ConsenSys/scribble/blob/d4da2edac356793067afb36dd2588d4f91746a84/src/spec-lang/tc/internal_types/builtin_struct_type.ts#L3-L19)
- [`ImportRefType`](https://github.com/ConsenSys/scribble/blob/d4da2edac356793067afb36dd2588d4f91746a84/src/spec-lang/tc/internal_types/import_ref_type.ts#L4-L20)
- [Mapping of predefined builtins](https://github.com/ConsenSys/scribble/blob/d4da2edac356793067afb36dd2588d4f91746a84/src/spec-lang/tc/builtins.ts)

## Changes
- [x] Introduced `VersionDependentType` to indicate that certain type is only accessible in certain version range.
- [x] Introduced `getTypeForCompilerVersion()`.
- [x] Introduced `BuiltinStructType`.
  - Strictened type requirements to use `VersionDependentType` and explicitly indicate version range when members are accessible.
- [x] Introduced `ImportRefType`.
- [x] Introduced list of predefined builtins in `src/types/builtins.ts`.
  - Strictened type requirements to use `VersionDependentType` and explicitly indicate version range when members are accessible.
  - Explicitly marked what builtins are still missing and what documentation to dig for details.
- [x] Introduced related tests.

Regards.


src/types/builtins.ts
```diff
@@ -0,0 +1,246 @@
...
+
+export const BuiltinAddressMembers = new Map<string, VersionDependentType>([
+    ["balance", [new IntType(256, false), ">=0.4.13"]],
+    [
+        "staticcall",
+        [
+            new FunctionType(
+                undefined,
+                [new PointerType(new BytesType(), DataLocation.Memory)],
+                [new BoolType(), new PointerType(new BytesType(), DataLocation.Memory)],
```
cd1m0, line 231: The return type as specified here (`(bool, bytes memory)`) only applies after 0.5.0. Before 0.5.0 it was just `bool`. So we will need a way to specify multiple different types, depending on the version.

cd1m0 (changes requested):
Looks good. Just one comment to consider before merging

src/types/builtins.ts
cd1m0: After discussion with @blitz-1306  we decided to fix this as a later PR

cd1m0 (approved):
LGTM. As discussed we will support different version-dependent types for the same name in a future PR.

PR #64, opened 2021-09-17
Support composing getter/accessor types

blitz-1306:
## Premise
This PR is a part of logic to address https://github.com/ConsenSys/scribble/issues/75.

## Changes
- [x] Tweaks for `assert()` to support placeholders. This would allow to delay heavy logic (like `ASTNode.print()` or `TypeNode.pp()`) until (and if) the assertion is failed and there is a real necessity to throw an error (with composed message).
- [x] Introduced `typeNameToSpecializedTypeNode()` shorthand to regularly used `specializeType(typeNameToTypeNode(node), loc)`.
- [x] Introduced `enumToIntType()`  to convert enum definition types to `int` value types.
- [x] Introduced `getterArgsAndReturn()` and `getterTypeForVar()` to implement requested functionality.
- [x] Tweaked `EnumDefinition.toUintTypeString()` to use `enumToIntType()`.
- [x] Other performance and readability tweaks in type-related utils.
- [x] Added related tests.

Regards.


src/types/utils.ts
```diff
@@ -257,3 +245,152 @@ export function typeNameToTypeNode(astT: TypeName): TypeNode {
...
+ * 2. Enum definitions turned to uint of minimal fitting size.
+ * 3. Struct definitions turned to tuples.
+ *   3.1. Mappings are completely ommited.
+ *   3.2. Arrays of root structs are ommited, while preserved for nested structs.
+ * 4. Data locations are skipped.
+ */
+function toValueType(type: TypeNode, depth = 0): TypeNode {
+    if (type instanceof PointerType) {
+        return toValueType(type.to, depth);
+    }
```
blitz-1306, line 301: @cd1m0 I wonder if this is valid. Maybe we should skip `specializeType()` in downstream instead.

test/samples/solidity/getters.sol
```diff
@@ -0,0 +1,45 @@
...
+    struct S2 {
+        uint[] a;
+        address b;
+    }
+
+    struct S {
+        uint[] a;
+        int8 b;
+        string x;
+        S1 s1;
```
blitz-1306, line 25: @cd1m0 This is very interesting edge-case. Nested tuples preserve arrays, while root structs do not.

src/types/utils.ts
cd1m0: I would prefer to generalize types here, and keep types "specialized" everywhere else. Both maru-ir and scribble depend on `variableDeclarationToTypeNode` to return a specialized type, and they specialize the return of `typeNameToTypeNode` whenever they use it. It seems that we only generalize types when we use them for something web3 related

test/samples/solidity/getters.sol
cd1m0: Yep. I also noticed this one

src/types/utils.ts
```diff
@@ -257,3 +245,152 @@ export function typeNameToTypeNode(astT: TypeName): TypeNode {
...
+    );
+
+    while (true) {
+        if (type instanceof ArrayTypeName) {
+            argTypes.push(new IntType(256, false));
+
+            type = type.vBaseType;
+        } else if (type instanceof Mapping) {
+            argTypes.push(
+                toValueType(typeNameToSpecializedTypeNode(type.vKeyType, DataLocation.Memory))
```
cd1m0, line 369: Why are we calling `toValueType` here? I would think that if we are computing the type for a getter for var `mapping(string => uint)` the argument would be `string memory` not `string`?

cd1m0 (changes requested):
One question. Otherwise lgtm

src/types/utils.ts
blitz-1306: Due to `Enum`s possible as mapping keys.

blitz-1306: Regarding `string memory`, see https://github.com/ConsenSys/solc-typed-ast/commit/83cda55cac6638e4dd593bf56869eb33b22584d4. Allowed non-definition types to be pointers. Also pointers are reconstructed to point to `memory` data location.

src/types/utils.ts
```diff
@@ -269,3 +262,179 @@ export function typeNameToTypeNode(astT: TypeName): TypeNode {
...
+
+    return new IntType(size, false);
+}
+
+/**
+ * Converts `type` to value type. Following rules are applied:
+ * 1. Contract definitions turned to address.
+ * 2. Enum definitions turned to uint of minimal fitting size.
+ */
+function toValueType(type: TypeNode, encoderVersion: ABIEncoderVersion): TypeNode {
```
cd1m0, line 311: The name of the function is a bit general. Maybe something like `toAccessorType`? Not sure. Also not a blocker, might merge as is.

test/samples/solidity/getters_07.sol
```diff
@@ -0,0 +1,29 @@
+pragma solidity ^0.7.0;
```
cd1m0, line 1: As a regression test we might want to also add an ABIEncoderV1 test. Not a blocker

cd1m0 (approved):
LGTM. Minor nits that can be fixed in a followup PR

cd1m0 (changes requested):
So after playing around I found another issue. So there is a difference between what the signature of accessors are at the _Solidity_ level, and at the level of the contract's _ABI_. Specifically contracts returned from public vars are not treated as addresses at the solidity level, just at the API level. The same is true for enums - at the solidity levels they are treated as enums, not uints. This samples shows the failures:

```solidity
pragma solidity ^0.8.0;
contract T {}

contract Foo {
    enum E {A,B,C}
    
    T public t;
    E public e;
    
    function main() public {
        address a = this.t();
        uint8 b = this.e();
    }
}
```

Both assignments in main will show type errors. One way to fix this is to:

1. Change `getterArgsAndReturn` to not use `toValueType`, and instead just convert non-memory pointer types to memory pointers
2. Add another `getterExternalArgsAndReturn` which calls `getterArgsAndReturn`, and calls `toValueType` on it.

Also we have code duplication with `VariableDeclaration.canonicalSignatureType` and `VariableDeclaration.getterCanonicalSignature`. First of all, not sure why we have two separate functions? I think they should do the same. And second, I think they should just call `getterExternalArgsAndReturn`

blitz-1306:
@cd1m0 I guess we somehwere went too far with public getters on ABI | internal level. Instead of complicating the code base I think we can remove `toValueType()` and dependent parts, just leave the `getterTypeForVar()` and `getterArgsAndReturn()` to build internal getter function type. Then adding a logic to convert it to ABI representation should be easier and may be left to consumers (or maybe done as a followup). I also changed approach for the tests to replicate actual function signatures (it looks more transparent). See https://github.com/ConsenSys/solc-typed-ast/pull/64/commits/bf45d33711e23ef99727b13d3141fb6bcd8d7559.

test/samples/solidity/getters_07.sol
cd1m0: Its added in my followup PR so no need to do this here.

cd1m0 (approved):
LGTM!

src/types/utils.ts
cd1m0: Irrelevant since bf45d33

PR #66, opened 2021-09-20
Add support for user-passed compiler settings

### LINKED ISSUE #65: Support user-specified compiler settings
## Description

We want to allow the user to pass arbitrary JSON to be used in the `settings: {...}` section of the standard input. This requires adding an optional argument to all `compile*` API entry points, and a new command-line option `compiler-settings` added to `sol-ast-compile`

## Context

This is important for cases where the code may not even compile without special settings (e.g. sometimes you need `optimizer: {enabled: true, ..}` to compile, otherwise you violate stack restrictions).

## Tests [optional]

Make sure the new option is supported and reflected in the help message

cd1m0:
This PR fixes #65 .

Namely it:

- adds a new optional argument `compilerSettings` to all compile API entry points (`compile`, `compileSol`, `compileJSON`, `compileSourceString` and `compileJSONData`) that should be a valid JSON object, and gets merged with the `setttings` field of the standard compiler input

- adds a new `--compiler-settings` command line option to the `sol-ast-compile` tool, that when present should contain a valid JSON that gets passed in as compiler settings

### DIFF
--- a/src/bin/compile.ts
+++ b/src/bin/compile.ts
@@ -41,7 +41,7 @@ const cli = {
         "source"
     ],
     number: ["depth"],
-    string: ["mode", "compiler-version", "path-remapping", "xpath"],
+    string: ["mode", "compiler-version", "path-remapping", "xpath", "compiler-settings"],
     default: {
         depth: Number.MAX_SAFE_INTEGER,
         mode: modes[0],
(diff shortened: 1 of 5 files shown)


blitz-1306 (approved):
Looks good.

PR #68, opened 2021-09-21
Modify compile* endpoints to support selecting compilation output

cd1m0:
This PR adds the capability to specify what specific output we want when compiling, by adding an extra optional `compilationOutput` argument to the API entrypoints `compile`, `compileSol`, `compileJSON`, `compileSourceString` and `compileJsonData`. `compilationOutput` is an array of values from the enum `CompilationOutput`, which correspond to the possible options to be specified in the `outputSelection` field of the standard input json (see [here](https://docs.soliditylang.org/en/v0.8.0/using-the-compiler.html#input-description) for more details.

### DIFF
--- a/src/ast/constants.ts
+++ b/src/ast/constants.ts
@@ -89,6 +89,32 @@ export enum TimeUnit {
     Years = "years"
 }
 
+export enum CompilationOutput {
+    AST = "ast",
+    ABI = "abi",
+    DEVDOC = "devdoc",
+    USERDOC = "userdoc",
+    METADATA = "metadata",
+    IR = "ir",
+    IR_OPTIMIZED = "irOptimized",
+    STORAGE_LAYOUT = "storageLayout",
+    EVM = "evm",
+    EVM_ASSEMBLY = "evm.assembly",
+    EVM_LEGACY_ASSEMBLY = "evm.legacyAssembly",
+    EVM_BYTECODE = "evm.bytecode",
+    EVM_BYTECODE_OBJECT = "evm.bytecode.object",
+    EVM_BYTECODE_OPCODES = "evm.bytecode.opcodes",
+    EVM_BYTECODE_SOURCEMAP = "evm.bytecode.sourceMap",
+    EVM_BYTECODE_LINKREFERENCES = "evm.bytecode.linkReferences",
+    EVM_BYTECODE_GENERATEDSOURCES = "evm.bytecode.generatedSources",
+    EVM_DEPLOYEDBYTECODE_IMMUTABLEREFERENCES = "evm.deployedBytecode.immutableReferences",
+    EVM_METHODIDENTIFIERS = "evm.methodIdentifiers",
+    EVM_GASESTIMATES = "evm.gasEstimates",
+    EWASM_WAST = "ewasm.wast",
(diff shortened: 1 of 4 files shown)


src/compile/utils.ts
```diff
@@ -342,6 +354,7 @@ export function compileSourceString(
     sourceCode: string,
     version: string | CompilerVersionSelectionStrategy,
     remapping: string[],
+    compilationOutput: CompilationOutput[] = [CompilationOutput.ALL],
```
blitz-1306, line 357: If we would move `compilationOutput` after `compilerSettings`, then there will be less compatibility breaks. Other modified functions are internal, so they not cause breaks. If we swap these arguments, we would be able to do minor version bump instead of major.

blitz-1306 (commented):
Good job. Few things to consider.

src/ast/constants.ts
```diff
@@ -89,6 +89,32 @@ export enum TimeUnit {
     Years = "years"
 }
 
+export enum CompilationOutput {
```
blitz-1306, line 92: Would be cool to have Solidity reference documentation entry for this enum. Also, this file is for AST-related constants. Compile-related constants better to be stored in `src/compile/constants.ts`.

src/compile/utils.ts
```diff
@@ -81,64 +90,61 @@ function mergeCompilerSettings<T extends Solc04Input | Solc05Input>(input: T, se
...
+    };
+
+    let inp: Solc04Input | Solc05Input;
+
+    if (lt(version, "0.5.0")) {
+        partialInp.sources = {
+            [fileName]: content
+        };
+
+        inp = partialInp as Solc04Input;
```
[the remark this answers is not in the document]
cd1m0: Fixed in 125b9f3

src/compile/utils.ts
```diff
@@ -55,16 +56,24 @@ export function getCompilerForVersion(version: string): any {
     );
 }
 
+type PartialSolcInput = {
+    language: "Solidity";
+    settings: { remappings: string[]; outputSelection: any; [otherKeys: string]: any };
+    [otherKeys: string]: any;
+};
```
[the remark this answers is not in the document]
cd1m0: Fixed in 125b9f3

src/ast/constants.ts
cd1m0: Fixed in 125b9f3

src/compile/utils.ts
cd1m0: On second thought I would actually like to keep them in this order. My motivation is:

1. `compilationOutput` and `compilerSettings` are introduced almost at the same time (they will differ only be 1 patch version in solc-typed-ast and scribble). So I don't think there will be consumers that know about one, and not the other

2. I think consumers will specify  `compilationOutput` more often than `compilerSettings`. So I would rather keep that one first, so that they don't have to insert 'undefined'.

3. Since `compilationOutput` is of type array, and `compilerSettings` is of type any (but is usualy an object), with the current order, if a user forgets to put in `compilationOutput` but instead directly passes `compilerSettings`, they will get a compiler error. If we switch the order they will not get a compiler error.

blitz-1306 (approved):
LGTM

PR #71, opened 2021-09-28
Pretty printing

blitz-1306:
## Dependence
- Requires #64 to be merged first.

## Changes
- [x] Introduced `PPAble`, `PPish` types and `pp()` function for pretty-printing primitive and complex values.
- [x] Tweaked `details` argument of `assert()` to accept `PPish` array.
- [x] Added related tests.

Regards.

### DIFF
--- a/src/misc/pretty_printing.ts
+++ b/src/misc/pretty_printing.ts
@@ -4,7 +4,19 @@ export interface PPAble {
     pp(): string;
 }
 
-export type PPIsh = PPAble | ASTNode | string | number | boolean | bigint | null | undefined;
+export type PPIsh =
+    | PPAble
+    | ASTNode
+    | string
+    | number
+    | boolean
+    | bigint
+    | null
+    | undefined
+    | PPIsh[]
+    | Set<PPIsh>
+    | Map<PPIsh, PPIsh>
+    | Iterable<PPIsh>;
 
 export function isPPAble(value: any): value is PPAble {
     return value ? typeof value.pp === "function" : false;
@@ -12,7 +24,7 @@ export function isPPAble(value: any): value is PPAble {
 
 export function pp(value: PPIsh): string {
     if (value instanceof ASTNode) {
-        return value.type + "#" + value.id;
+        return value.type + " #" + value.id;
     }
 
     if (value === undefined) {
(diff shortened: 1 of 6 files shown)


src/misc/pretty_printing.ts
```diff
@@ -0,0 +1,71 @@
...
+        typeof value === "boolean" ||
+        typeof value === "bigint"
+    ) {
+        return String(value);
+    }
+
+    if (isPPAble(value)) {
+        return value.pp();
+    }
+
```
cd1m0, line 35: Could we add the following ifs here (to allow for pp() to handle (potentially nested) arrays, maps and iterables):

```
if (value instanceof Array) {
  return ppArr(value);
}

if (<value is an Iterable>) {
   return ppIter(value);
}

if (value instanceof Set) {
   return ppSet(value);
}

if (value instanceof Map) {
   return ppMap(value);
}
```

After we add this, we should add tests that we can pretty-print types like `string[][]`, `Map<string, Map<number, number>>`, `Set<string>[][]` etc...

cd1m0 (changes requested):
Just 1 comment :)

src/misc/utils.ts
```diff
@@ -29,13 +29,13 @@ export function assert(
             let part: string;
 
             if (detail instanceof ASTNode) {
-                part = detail.type + " #" + detail.id;
```
cd1m0, line 32: This if case can be removed (its already handled in pp()).

src/misc/utils.ts
```diff
@@ -29,13 +29,13 @@ export function assert(
...
+                part = pp(detail);
 
                 nodes.push(detail);
-            } else if (detail instanceof TypeNode) {
-                part = detail.pp();
+            } else if (detail instanceof Array) {
```
cd1m0, line 35: This case can be removed after we add the additional ifs in pp().

cd1m0 (changes requested):

src/misc/utils.ts
blitz-1306: Addressed in https://github.com/ConsenSys/solc-typed-ast/commit/6e612113efada422eb324ed77841541855af8c10

src/misc/utils.ts
blitz-1306: Addressed in https://github.com/ConsenSys/solc-typed-ast/commit/6e612113efada422eb324ed77841541855af8c10

src/misc/pretty_printing.ts
blitz-1306: Addressed in https://github.com/ConsenSys/solc-typed-ast/commit/6e612113efada422eb324ed77841541855af8c10.

Still, we should be careful with possible recursion here.

cd1m0 (approved):
LGTM

PR #76, opened 2021-10-04
Support for Solidity 0.8.8, 0.8.9 and 0.8.10

blitz-1306:
## Changes
- [x] Updated dependencies: `web3-eth-abi` to `^1.6.0`, also development dependencies to work with latest development tools.
- Added support of Solidity 0.8.8, 0.8.9 and 0.8.10 on compiler, AST levels and type system levels:
  - [x] Introduced AST node `UserDefinedValueTypeDefinition` and related modern (compact) AST processor.
  - [x] Tweaked `ContractDefinition` and `SourceUnit` AST nodes to have `vUserDefinedValueTypes()` getter and reference scoped `UserDefinedValueTypeDefinition`.
  - [x] Tweaked `ExportedSymbol` to also mention `UserDefinedValueTypeDefinition` (can be exported as it is a part of source unit).
  - [x] Added support for `UserDefinedValueTypeDefinition` in `ASTNodeFactory`.
  - [x] Added support for `UserDefinedValueTypeDefinition` on type system level.
  - [x] Tweaked AST-to-source writing components to respect `UserDefinedValueTypeDefinition`.
  - [x] Updated NODE_LIST doc to reference `UserDefinedValueTypeDefinition` node.

## Notes
Changes of `ExportedSymbol` are breaking, but not changing API way too much. So we can bump only minor version.

## Related links
- [Solidity 0.8.8 release notes](https://github.com/ethereum/solidity/releases/tag/v0.8.8)
- [Solidity 0.8.8 release blog post](https://blog.soliditylang.org/2021/09/27/solidity-0.8.8-release-announcement/)
- [Solidity 0.8.9 release notes](https://github.com/ethereum/solidity/releases/tag/v0.8.9)
- [Solidity 0.8.9 release blog post](https://blog.soliditylang.org/2021/09/29/solidity-0.8.9-release-announcement/)
- [User Defined Value Types in Solidity](https://blog.soliditylang.org/2021/09/27/user-defined-value-types/)
- [Solidity 0.8.10 release notes](https://github.com/ethereum/solidity/releases/tag/v0.8.10)
- [Solidity 0.8.10 release blog post](https://blog.soliditylang.org/2021/11/09/solidity-0.8.10-release-announcement/)

Regards.


src/ast/implementation/declaration/user_defined_value_type_definition.ts
```diff
@@ -0,0 +1,58 @@
...
+    /**
+     * The name of the user-defined value type definition
+     */
+    name: string;
+
+    /**
+     * Canonical name (or qualified name), e.g. `DefiningContract.SomeType`.
+     * Is `undefined` for Solidity 0.8.8. Available since Solidity 0.8.9.
+     */
+    canonicalName?: string;
```
cd1m0, line 16: Maybe in a compute this property for 0.8.8 so we can always assume its present?

cd1m0 (approved):
Awesome work! Good to merge. One nit - but its not a blocker. Your call if you want to fix it or leave it as is

src/ast/implementation/declaration/user_defined_value_type_definition.ts
blitz-1306: Does not seem too complex to fix in-place. Will look into.

blitz-1306: The thing here is that this node is constructed before "access to parent" is available. So it would require a postprocessing pass to fix this. I'm arguing if it would be easier to just drop `canonicalName` and make it a getter instead due to field value should change if parent changes. Currently it is just static value from compiler itself and does not react on tree modification. However, this would involve much greater cleanup, as `canonicalName` is available in few other type nodes (structs, enums and whatever else). So lets keep current state of things for now, then do a cleanup pass to compyte such things on runtime.

blitz-1306: Created issue #80 to track this.

PR #82, opened 2021-11-15
Introduce `canonicalName` as getter, support `UserDefinedValueTypeDefinition` in `ASTNodeFactory.makeIdentifierFor()`

### LINKED ISSUE #80: Introduce `canonicalName` property as getter
## Description

Currently few type name nodes have a `canonicalName` property, that just stores vaule since compiler AST reading. If the parent node is changed, the `canonicalName` for it's child nodes is not updated properly. It would be better if `canonicalName` is introduced as getter  to be able to make parent lookups and compute values on demand. Compiler values may be a good input to use for the tests.

blitz-1306:
## Tickets
- Closes #80 

## Changes
- [x] Introduced `canonicalName` as getter for `EnumDefinition`, `StructDefinition` and `UserDefinedValueTypeDefinition`.
- [x]  Added `UserDefinedValueTypeDefinition` as a `target` type for `ASTNodeFactory.makeIdentifierFor()`.
- [x] Tweaked affected tests.

## Notes
There are constructor signature changes for `EnumDefinition`, `StructDefinition` and `UserDefinedValueTypeDefinition` as the `canonicalName` is no longer passed as an argument. This would break backward compatibility.

Regards.

### DIFF
--- a/src/ast/ast_node_factory.ts
+++ b/src/ast/ast_node_factory.ts
@@ -110,7 +110,6 @@ const argExtractionMapping = new Map<ASTNodeConstructor<ASTNode>, (node: any) =>
         EnumDefinition,
         (node: EnumDefinition): Specific<ConstructorParameters<typeof EnumDefinition>> => [
             node.name,
-            node.canonicalName,
             node.vMembers,
             node.nameLocation,
             node.raw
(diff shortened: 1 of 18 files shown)


cd1m0 (approved):
lgtm

PR #88, opened 2022-01-13
Re-factor of compile* logic and other cleanup

cd1m0:
This PR includes:

1. Re-factors the compile logic in all the compile* entry points to:
   a) first call `findAllFiles` to identify all imports
   b) build correct standard input JSON with all needed files
   c) call the underlying compiler. (without specifying import callbacks for WASM)
   
2. Changes the signature of compile() to accept a map from file name -> contents instead of a single file

3. Various fixes to get a clean test run with native compilers

4. Fixes for first round of review on the native_compilers branch.

Several things are still missing/need attention:

1. This doesn't include a fix for  the rename TopLevelNodes => FileLevelNodes

2. If you switch back the default execution to WASM several tests fail. The failures seem to be for 2 reasons:

 - the ids of some nodes are different
 - the standard JSON output format for 0.4.x seems to be different between WASM and native. The native output format is better (its the same as 0.5.x..)
 
 3. We should make all the mocha tests parametric in the front end - so that we can run either with just a flip of the switch. 
 
 4. We still need a unit test that checks that for all samples we have, the two different compilers produce the SAME ast
 
 5. There are various TODOs in the code that need attention


src/compile/frontends/native_compilers.ts
```diff
@@ -116,19 +118,16 @@ export async function httpsGetSync(url: string): Promise<Buffer> {
...
-
-    return await reqPromise;
 }
 
 async function getCompilerMDForPlatform(prefix: string): Promise<CompilerPlatformMetadata> {
     const cachedListPath = path.join(CACHE_DIR, prefix, "list.json");
 
     if (fse.existsSync(cachedListPath)) {
-        //console.error(`Hit cache for list ${prefix}`);
         return fse.readJSONSync(cachedListPath) as CompilerPlatformMetadata;
```
blitz-1306, line 127: We may need to have a logic for recheking, as the list is updated after some time.

src/compile/frontends/wasm.ts
```diff
@@ -1,21 +1,43 @@
...
-        this.module = require(this.moduleName);
     }
 
-    abstract compile(inputJSON: any, finder: ImportFinder): any;
+    public get module(): any {
+        if (this._module === undefined) {
+            this._module = require(`solc-${this.version}`);
+        }
+
+        return this._module;
```
blitz-1306, line 26: We can rely on just following:
```js
return require(`solc-${this.version}`);
```

src/compile/inference/imports.ts
```diff
@@ -0,0 +1,88 @@
...
+            (tld) => tld.kind === TopLevelNodeKind.Import
+        ) as TLImportDirective[];
+
+        for (const imp of imports) {
+            const path = imp.path;
+            if (path.startsWith("./") || path.startsWith("../")) {
+                const importingFileDir = dirname(filePath);
+                // TODO(dimo): We have to check that the behavior of normalize(join(...)) below
+                // matches the behavior described in this section https://docs.soliditylang.org/en/v0.8.8/path-resolution.html#relative-imports
+                // and that it works for all compiler versions
```
blitz-1306, line 79: This seems like a unit-test case. We may check it in out tests.

blitz-1306 (approved):
This is good enougn to merge with the other WIP branch. We can address remarks and do tweaks on a target branch.

sambacha:
Would it be possible to have a fallback to the github releases for fetching binaries? soliditylang.org has had outages recently, frustrating to have a CI process fail because it couldn't get the binaries

blitz-1306:
@sambacha We will rechek in the other PR #87 and do some other polishing. Thanks.

cd1m0:
@sambacha If you specify "--compiler-kind wasm" on the command line it won't rely on fetching the binaries (it currently still pre-installs all the npmjs solidiity modules... which is a problem in itself...). Is this sufficient for you?

cd1m0:
@sambacha sorry I didn't read your question closely enough. I think this is a good idea!

cd1m0:
@sambacha for now we decided to leave the WASM backend as the default. So unless you change your code to explicitly use native compilers, your CI shouldn't be impacted at all. 

If you do want to test native compilers in CI, you can follow our plan - we are going to try and add a mounted volume with the binaries pre-downloaded to avoid failures, and speed-up the CI. When we get it set up we can share with you the steps if you want to replicate in your CI setup?

PR #92, opened 2022-02-09
Dowloading WASM compilers on-demand instead of preinstalling from NPM

blitz-1306:
## Preface
Currently we preinstall a lot of WASM compilers of Solidity from NPM. Recent introduction of native compilers dowloading on-demand allows us to start figuring out if we are able to do the same for WASM compilers. This PR is a PoC to show the problems and answer multiple "why didn't we introduced this earlier?".

## Changes
- [x] Removed multiple NPM dependencies of Solc from `package.json`. Now only latest is used.
- [x] Merged `native.ts` and `wasm.ts` modules to `compiler.ts` module. `WasmCompiler` were simplified, as now it have to load WASM/BIN compiler instance and setup Solc methods before invoking the compiler.
- [x] Renamed `getNativeCompilerForVersion()` to `getCompilerForVersion()`. Added `kind` parameter and did related tweaks to allow downloading WASM from Solidity builds historical archive.
- [x] Tweaked pre-downloading steps in CI config.
- [x] Added support of Solidity 0.8.12 on compiler level. There are no changes on AST level for this release of compiler.

## Related links
- [Solidity 0.8.12 release notes](https://github.com/ethereum/solidity/releases/tag/v0.8.12)
- [Solidity 0.8.12 release blog post](https://blog.soliditylang.org/2022/02/16/solidity-0.8.12-release-announcement/)

## Notes & Concerns
- https://github.com/ethereum/solidity/issues/10329. I bumped the discussion there. As a workaround we will switch to use `bin` distribution for some time - it should be the same as published in NPM.
  - Switched to `bin` release instead of `wasm`, as there is 0.5.17 release available.
- Due to some reasons, we now encountering OOMs during compiler loading for the tests. The workaround may be to add tiny standalone executable script, that will load WASM, feed input and return compiler output. Seems to be not the best option, but guarantee proper memory deallocation, that may not happen if we try to load multiple WASM modules during a single process.
  - Switched to `bin` release instead of `wasm`, as this allows to avoid OOMs.
  - Also, running on **NodeJS v16.14.0** (`lts/gallium`) is not causing to OOMs.

Regards.


cameel:
> It appears that latest version of Solc package is not able to properly setup methods to legacy 0.4 WASM compiler instances, So we still using two of them.

Can you report this as a bug in [solc-js](https://github.com/ethereum/solc-js)? Or is it one of already reported problems? solc-js is generally meant to be an universal wrapper that can handle all past binaries.

> Due to some reasons, we now encountering OOMs during compiler loading for the tests.

I don't think the compiler binaries get unloaded from memory until your process ends so it might be due to this. Loading all of them simultaneously is a bit of an extreme use case :)

----

By the way, have you tried latest solc-js from develop? We started the process of migrating to Typescript and the structure changed a bit so I'd be interested to hear if you can use it without problems.

blitz-1306:
@cameel Thanks for the input, we will consider.

test/unit/compile/utils.spec.ts
```diff
@@ -2,29 +2,46 @@ import expect from "expect";
...
+                expect(e.message).toMatch("");
+            }
         });
 
         const strategy = new LatestAndFirstVersionInEachSeriesStrategy();
 
         for (const version of strategy.select()) {
-            it(`Compiler ${version} is accessible`, () => {
-                expect(getWasmCompilerForVersion(version)).toBeInstanceOf(WasmCompiler);
+            it(`Compiler ${version} is accessible`, async () => {
```
cd1m0, line 41: In this test can we also check that `getCompilerForVersion(version, CompilerKind.WASM)`is also correct?

src/compile/kinds/compiler.ts
```diff
@@ -1,14 +1,155 @@
...
+        });
+    }
+}
+
+export class WasmCompiler extends Compiler {
+    async compile(input: SolcInput): Promise<any> {
+        const module = require(this.path);
+
+        if (satisfies(this.version, "^0.4.13")) {
+            const wrappedModule = solc0426.setupMethods(module);
```
cd1m0, line 79: It sucks that we have to keep a dependency on 0.4.13 just for the setupMethods call, since that will still keep those vulnerable dependencies that dependbot complains about. I wonder if there is some other workaround?

cd1m0 (commented):
Lgtm honestly. We just gotta figure out solutions to the issues you pointed out in your description.

src/compile/kinds/compiler.ts
blitz-1306: As suggested [above](https://github.com/ConsenSys/solc-typed-ast/pull/92#issuecomment-1034976261) we can consider to mitigate via opening an issue in SolcJS repo. I will try to make minimal reproducible example  to see if this is really an issue.

blitz-1306: This is actuall a bug. I didn't updated input composition to use logic for Solc 0.5.* and above.

blitz-1306: Addressed in https://github.com/ConsenSys/solc-typed-ast/pull/92/commits/243a46a7b1732f537208d3f22a4c0c00c5729d0e

src/compile/kinds/compiler.ts
```diff
@@ -1,14 +1,152 @@
...
+    async compile(input: SolcInput): Promise<any> {
+        const module = require(this.path);
+        const wrappedModule = solc.setupMethods(module);
+        const output = wrappedModule.compile(JSON.stringify(input));
+
+        return JSON.parse(output);
+    }
+}
+
+type CompilerMapping = [CompilerKind.Native, NativeCompiler] | [CompilerKind.WASM, WasmCompiler];
```
cd1m0, line 82: Nice! Didnt know you could do this.

test/integration/sourcemap/solc_comparison.spec.ts
```diff
@@ -74,93 +74,94 @@ const samples = [
...
+                            continue;
+                        }
+
+                        const solcCoords = solcStart + ":" + solcLen;
+                        const compCoords = compStart + ":" + compLen;
+
+                        const solcFragment = getSourceFragment(solcStart, solcLen, writtenSource);
+                        const compFragment = getSourceFragment(compStart, compLen, writtenSource);
+
+                        if (compFragment === solcFragment && compCoords === solcCoords) {
```
cd1m0, line 129: This is not a blocker, but the way this is written its not optimal. We only need `solcFragment` and `compFragment` if  `compCoords !== solcCoords`. So lines 123-131 can be optimized to:

```
if (solcStart === compStart && solcLen === compLen) {
    continue;
}

const solcCoords = solcStart + ":" + solcLen;
const compCoords = compStart + ":" + compLen;

const solcFragment = getSourceFragment(solcStart, solcLen, writtenSource);
const compFragment = getSourceFragment(compStart, compLen, writtenSource);
```

cd1m0 (approved):
LGTM

src/compile/kinds/compiler.ts
```diff
@@ -1,14 +1,152 @@
...
+        prefix = getCompilerPrefixForOs();
+    } else if (kind === CompilerKind.WASM) {
+        /**
+         * Using BIN distribution here due to WASM distributions do not have 0.5.17 build and also OOM issues.
+         *
+         * @see https://github.com/ethereum/solidity/issues/10329
+         *
+         * @todo Reconsider this at some point.
+         */
+        prefix = "bin";
```
cameel, line 105: Just FYI, the wasm binary is now available: https://github.com/ethereum/solc-bin/blob/gh-pages/wasm/soljson-v0.5.17+Bcommit.d19bba13.js

blitz-1306: FYI, OOMs are happening on NodeJS 12 and not happening on NodeJS v16.14.0 (`lts/gallium`). Just in case if you will encounter similar question.

PR #93, opened 2022-02-10
Cleanup import resolution logic

cd1m0:
As of #87 we changed solc-typed-ast to infer the set of all imported files, and pass them in to the compiler in the standard JSON. The idea was that we could support a uniform interface between wasm and native compilers - since one of them (wasm) has support for resolving callbacks, but the native one doesn't. Also this way compilers don't need to look around the file system or do anything else - we give them everything they need to compile.

The logic of finding all of the imports however is a little more complex as described [here](https://docs.soliditylang.org/en/v0.8.8/path-resolution.html). Our (ok, ok my) original code was a little hacky. In this PR we clean it up to follow more precisely the algorithm described in the [docs](https://docs.soliditylang.org/en/v0.8.8/path-resolution.html).

Specifically:
1. The keys in the map `files` computed in `findAllFiles` are now proper **source unit names** as defined in docs.
2. `findAllFiles`  was cleaned up a little ,and now no-longer returns a new map.
3. We pass the original remappings to the underlying compilers again (this is needed so that the underlying compiler computes the same source unit names as we do, and finds them in the files map)

### DIFF
--- a/src/compile/inference/imports.ts
+++ b/src/compile/inference/imports.ts
@@ -98,12 +98,7 @@ export function findAllFiles(
     resolvers: ImportResolver[]
 ): void {
     // Queue of source unit names to process
-    const queue: string[] = [];
-
-    for (const fileName of files.keys()) {
-        queue.push(fileName);
-    }
-
+    const queue: string[] = [...files.keys()];
     const visited = new Set<string>();
 
     while (queue.length > 0) {
--- a/src/compile/inference/imports.ts
+++ b/src/compile/inference/imports.ts
@@ -1,6 +1,7 @@
 import fse from "fs-extra";
-import { dirname, isAbsolute, join, normalize } from "path";
+import { dirname, normalize } from "path";
 import { CompileInferenceError, ImportResolver, Remapping } from "..";
+import { assert } from "../..";
 import {
     AnyFileLevelNode,
     FileLevelNodeKind,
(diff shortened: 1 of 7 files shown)


src/compile/inference/imports.ts
```diff
@@ -38,66 +33,97 @@ function isPathWithRelativePrefix(path: string): boolean {
...
-): Map<string, string> {
-    const queue: Array<[string | undefined, string]> = [];
+): void {
+    // Queue of source unit names to process
+    const queue: string[] = [];
 
     for (const fileName of files.keys()) {
-        queue.push([undefined, fileName]);
+        queue.push(fileName);
     }
```
blitz-1306, line 105: Wondering if spread will work here:
```typescript
const queue = [...files.keys()];
```

src/compile/inference/imports.ts
```diff
@@ -137,10 +159,8 @@ export function findAllFiles(
 
         for (const fld of flds) {
             if (fld.kind === FileLevelNodeKind.Import) {
-                queue.push([realPath, fld.path]);
+                queue.push(computeSourceUnitName(sourceUnitName, fld.path, remappings));
```
blitz-1306, line 162: **This is a good call**. I didn't figured it out during the last optimization pass and decided to preserve importer/imported paths as separate entities. But now I see that it caused way too much complications in codebase above.

blitz-1306 (approved):
I approve. One small question, that is not a blocker and not affect approach overall.

src/compile/inference/imports.ts
cd1m0: Fixed in [eda0998](https://github.com/ConsenSys/solc-typed-ast/pull/93/commits/eda09983d90f42073d7083baf8f79bd68802792f)

blitz-1306 (approved):
Looks good

PR #95, opened 2022-02-10
Predownload native compilers in CI

blitz-1306:
## Preface
This PR tweaks CI to use compiler downloading script prior to executing tests. It will help us to avoid `ETXTBSY` exit codes, probably caused by attempts to access compilers that are still downloading. We can reconfigure to use docker image later, when we decide where to publish it.

## Changes
- [x] Added CI commands to predownload native compilers and set ENV to use them instead of downloading on-demand.

## Notes
Part of the code, that is responsible for downloading on-demand will not be covered by the tests as the result of predownloading. Other option is to create a test that downloads only one compiler that is not used by other tests (or remove one rarely used compiler).

Regards.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -11,7 +11,10 @@ steps:
       environment:
           CODECOV_TOKEN:
               from_secret: CODECOV_TOKEN
+          SOL_AST_COMPILER_CACHE: /.compiler_cache
       commands:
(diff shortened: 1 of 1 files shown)


cd1m0 (approved):
lgtm!

PR #96, opened 2022-02-15
Fix off by one error in enumToIntType

swapnilraj:
The bug can be reproduced with this file https://github.com/ethereum/solidity/blob/develop/test/libsolidity/semanticTests/enums/enum_with_256_members.sol

Since the enum definition has 256 members it should fit in a uint8 but before the patch `enumToIntType` would return uint16

### DIFF
--- a/src/types/utils.ts
+++ b/src/types/utils.ts
@@ -284,7 +284,7 @@ export function enumToIntType(decl: EnumDefinition): IntType {
     let size: number | undefined;
 
     for (let n = 8; n <= 32; n += 8) {
-        if (length < 2 ** n) {
+        if (length <= 2 ** n) {
             size = n;
 
             break;


cd1m0:
Nice catch @swapnilraj! Will merge it shortly. Out of curiosity, did you run scribble against all of solc's conformance tests? Or did you catch this otherwise?

swapnilraj:
> Nice catch @swapnilraj! Will merge it shortly. Out of curiosity, did you run scribble against all of solc's conformance tests? Or did you catch this otherwise?

Hey @cd1m0, we are in the process of running a new version of [warp](https://github.com/NethermindEth/warp), against all the tests [here](https://github.com/ethereum/solidity/tree/develop/test/libsolidity/semanticTests). Will makes issues and/or PRs if we find something else as well.

cd1m0:
Thats awsome! Looking forward to any issues/prs!

cd1m0 (approved):
LGTM

PR #97, opened 2022-02-23
Migrate to NodeJS 16 LTS

blitz-1306:
## Changes
- [x] Migrated to NodeJS v16 LTS.
- [x] Switched from `bin` compiler distributions to `wasm` (as NodeJS v16 solves issues with OOMs during compiler runtime).
- [x] Introduced CLI option `--locate-compiler-cache`.
- [x] Updated README:
  - removed section about preinstalling 50+ compilers;
  - added section about manual invalidation of compiler cache;
  - other small tweaks and improvements.

Regards.


docker/download.sh
```diff
@@ -1,7 +1,11 @@
 #!/bin/bash
 
 ARCH="${1}"
-TARGET_DIR="/.compiler_cache/$ARCH"
```
cd1m0, line 4: Nit: Can we add a check on the number of arguments with an error message? Something like:

```
if [[ $# -ne 2 ]] ; then 
   echo "Usage: $0 ...."
   exit -1;
fi
```

src/compile/kinds/compiler.ts
```diff
@@ -128,16 +119,35 @@ export async function getCompilerForVersion<T extends CompilerMapping>(
...
+        assert(
+            digest === build.sha256,
+            `Downloaded ${prefix} compiler ${version} hash ${digest} does not match hash ${build.sha256} from "list.json"`
+        );
+
+        /**
+         * Native compilers are exeсutable files, so give them valid permissions.
+         * WASM compilers are loaded by NodeJS, so write them as common files.
+         */
+        const options = kind === CompilerKind.Native ? { mode: 0o555 } : undefined;
```
cd1m0, line 148: According to the writeFile [docs](https://nodejs.org/api/fs.html#fs_fs_writefile_file_data_options_callback) the default permissions for a new file are r/w - 0o666. Can we make it read-only (0o444) just in case?

cd1m0 (changes requested):
Awesome work as usual! Just 2 small nits

src/compile/kinds/compiler.ts
cd1m0: For wasm modules that is

blitz-1306: Addressed in https://github.com/ConsenSys/solc-typed-ast/pull/97/commits/bc60c2c4f19f074ef26fc2875d4e4b404b218c9f

docker/download.sh
blitz-1306: Addressed in https://github.com/ConsenSys/solc-typed-ast/pull/97/commits/bc60c2c4f19f074ef26fc2875d4e4b404b218c9f

PR #98, opened 2022-02-25
 Implement parser for rational literal types

JorikSchellekens:


src/types/ast/rational_literal.ts
```diff
@@ -0,0 +1,29 @@
+import { Range } from "../../misc";
+import { TypeNode } from "./type";
+
+interface Rational {
+  numerator: bigint;
+  denominator: bigint;
+}
+
+export class RationalLiteralType extends TypeNode {
+    public readonly literal?: Rational;
```
cd1m0, line 10: Question: Why is the literal member optional? Looking at the changes in  the grammar in typeString_grammar.pegjs literal is always provided.

yarn.lock
```diff
@@ -0,0 +1,3159 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
+# yarn lockfile v1
```
cd1m0, line 2: Would you mind removing this file from the commit?

cd1m0 (commented):
Thank you for your PR! Good work! Just 2 small nits in comments, and one small request. Could you add the following line:

```
     1.1;
```

After this line:

https://github.com/ConsenSys/solc-typed-ast/blob/master/test/samples/solidity/types/types.sol#L48

The idea is that if we have a rational literal in that file, then the following unit test will exercise the new code you added:

https://github.com/ConsenSys/solc-typed-ast/blob/master/test/unit/types/typestrings.spec.ts

blitz-1306 (changes requested):
Thanks for the contribution. I'm mostly +1 with the above comments from @cd1m0. If there would be any blockers for you to introduce the changes, just inform us. I will consider picking up the branch and doing the rest in that case.

src/types/ast/rational_literal.ts
JorikSchellekens: I had this same question about `IntLiteralType` and presumed there was some usecase I wasn't aware of.

JorikSchellekens:
> Thanks for the contribution. I'm mostly +1 with the above comments from @cd1m0. If there would be any blockers for you to introduce the changes, just inform us. I will consider picking up the branch and doing the rest in that case.

Thanks, should be good to go?

cd1m0 (approved):
lgtm

yarn.lock
blitz-1306: Addressed in https://github.com/ConsenSys/solc-typed-ast/pull/98/commits/19f312848b466885b4cdcb3fff2eb5dd7ec6c813

src/types/ast/rational_literal.ts
blitz-1306: Addressed in https://github.com/ConsenSys/solc-typed-ast/pull/98/commits/19f312848b466885b4cdcb3fff2eb5dd7ec6c813

blitz-1306 (approved):
Good to go. Thanks again for your efforts.

PR #104, opened 2022-03-14
Support function types in ABI-related logic

### LINKED ISSUE #103: Need to support function type names in `abi.ts`
Scribble received a [bug report](https://github.com/ConsenSys/scribble/issues/155) about an exception occuring when an annotated contract has function names in arguments/returns:

```
Error: Unexpected ABI Type function (address) returns (bool)
    at assert (/home/edkek/.nvm/versions/node/v10.24.1/lib/node_modules/eth-scribble/node_modules/solc-typed-ast/dist/misc/utils.js:36:11)
    at abiTypeToLibraryCanonicalName (/home/edkek/.nvm/versions/node/v10.24.1/lib/node_modules/eth-scribble/node_modules/solc-typed-ast/dist/types/abi.js:122:23)
    at args.vParameters.vParameters.map (/home/edkek/.nvm/versions/node/v10.24.1/lib/node_modules/eth-scribble/node_modules/solc-typed-ast/dist/ast/implementation/declaration/function_definition.js:61:102)
    at Array.map (<anonymous>)
    ....
```

The fundamental issue is that the `abiTypeToLibraryCanonicalName` function lacks support for FunctionTypes. We will likely have to add support for function names also to the `abiTypeToCanonicalName` and `toABIEncodedType` functions.

blitz-1306:
## Tasks
- Fixes #103

## Changes
- [x] Add support for function types when converting to ABI canonical names.

## Notes
- For ABI encoding the `function` type is replaced with `bytes24` - first 20 bytes for address of contract and 4 bytes for function selector. This happends for `external`/`public` functions with arguments of function type. See last entry of [the section for ABI Types in manual](https://docs.soliditylang.org/en/latest/abi-spec.html#types).
  - However, we are unable to just convert it to `bytes24` on-fly as word `function` is **used for function selector computation**. The replacement causes selector bytes mismatch.
- I'm not yet sure how this may affect resolution or overloading, especially for internal functions with such arguments. Compiler actually figures the difference of the function arg modifiers, so user can not pass just any function there. I'm wondering if there any edge-cases are involved and not yet figured out how to test this properly (for internal library functions in `abiTypeToLibraryCanonicalName()`).

Regards.

### DIFF
--- a/src/types/abi.ts
+++ b/src/types/abi.ts
@@ -12,6 +12,7 @@ import {
     BoolType,
     BytesType,
     FixedBytesType,
+    FunctionType,
     IntType,
     MappingType,
     StringType,
@@ -37,7 +38,9 @@ export const ABIEncoderVersions = new Set<string>([ABIEncoderVersion.V1, ABIEnco
  * 1. Contract definitions turned to address.
  * 2. Enum definitions turned to uint of minimal fitting size.
  * 3. Any storage pointer types are converted to memory pointer types.
- * 4. Throw an error on any nested mapping types
+ * 4. Throw an error on any nested mapping types.
+ *
+ * @see https://docs.soliditylang.org/en/latest/abi-spec.html
  */
 export function toABIEncodedType(
     type: TypeNode,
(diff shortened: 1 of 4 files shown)


cd1m0 (approved):
LGTM

cd1m0:
So for now I don't think we need to worry about internal functions, as the idea of a "function signature" I think is undefined for them. As far as I know its only defined for external/public functions, for which its used in the ABI. If we have upstream components that rely on signatures to resolve internal functions, then thats a bug in those components, and we should fix it.

PR #105, opened 2022-03-16
Support multiple file inputs to `compileSol()`

blitz-1306:
## Changes
- [x] Add overloads of `compileSol()` to enable support of compiling array of files via single compiler invocation. Added related tests.
- [x] Removed `resolveFiles()` as it more logical to use `findAllFiles()` and `parsePathRemapping()` separately. There is no real need to reparse remappings before each call to `findAllFiles()`.
- [x] The function `findAllFiles()` is now `async`. 
- [x] Minor smaller codebase improvements.

Regards.

### DIFF
--- a/src/compile/utils.ts
+++ b/src/compile/utils.ts
@@ -1,6 +1,7 @@
 import fse from "fs-extra";
 import path from "path";
-import { FileSystemResolver, getCompilerForVersion, ImportResolver, LocalNpmResolver } from ".";
+import { FileSystemResolver, getCompilerForVersion, LocalNpmResolver } from ".";
+import { assert } from "../misc";
 import {
     CompilerVersionSelectionStrategy,
     LatestVersionInEachSeriesStrategy,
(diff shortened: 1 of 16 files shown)


src/compile/utils.ts
```diff
@@ -212,21 +202,70 @@ export async function compileSol(
...
+    const npmResolver = new LocalNpmResolver();
+    const resolvers = [fsResolver, npmResolver];
+    const parsedRemapping = parsePathRemapping(remapping);
+    const files = new Map<string, string>();
+    const visited = new Set<string>();
+
+    for (const fileName of fileNames) {
+        const sourceCode = await fse.readFile(fileName, "utf-8");
+
+        npmResolver.baseDir = path.dirname(fileName);
```
cd1m0, line 239: This is not a blocker, but I am wondering if there is a situation, where different files might have different `node_modules` directories, and if in such a situation there might be breakage? I can't come up with a failure scenario, just something to consider I guess

cd1m0 (approved):
LGTM. Just one small nit/question we shuold quickly discuss. But if we can't come up with any failure scenarios for it then we merge :)

src/compile/utils.ts
blitz-1306: There may be a scenario when some absolute path was imported and it has own `node_modules`. I'm wondering the correct behavior in that case. What we currently do - is allow to resolve all files, relatively to each possible path. Otherwise it would be hard to pick common directory for different relative/absolute/virtual possible paths. I'm not sure about "correct"/reliable logic here.

cd1m0: I think the current behavior is as close as possible to "correct" in this case. If we run into such a case, and it causes problems lets revisit. For now I am down to merge

PR #107, opened 2022-03-21
Support for Solidity 0.8.13

blitz-1306:
## Changes
- [x] Updated dependencies: `solc` to `^0.8.13`, also development dependencies to work with latest development tools.
- [x] Added support of Solidity 0.8.13 on compiler and AST levels:
  - Added `flags` and `evmVersion` optional properties to `InlineAssembly` AST node.
  - Tweaked `UsingForDirective` AST node:
    - Property `vLibraryName` is now optional - it is missing when newly introduced tuple of functions is used.
    - Added `vFunctionList` and `isGlobal` optional properties.
    - Constructor signature been significantly changed.
  - Added `vUsingForDirectives` getter property for `SourceUnit` AST node to support file-level directives.
  - Added `FLUsingForDirective` node for file-level definitions parser with related grammar tweaks to properly support it.
  - Tweaked modern and legacy AST processors to respect the changes. Note that the legacy AST processor updates are maintaining backward compatibility only.
  - Tweaked `ASTNodeFactory` to respect the changes.
  - Tweaked AST-to-source writing components to respect the changes.
  - Note that `test/unit/ast/reading/meta/using_for.spec.ts` unit test is comletely deleted as it was obsolete and maintaining it was more costly than maintaining the integration tests for the `UsingForDirective`.

## Related links
- [Solidity 0.8.13 release notes](https://github.com/ethereum/solidity/releases/tag/v0.8.13)
- [Solidity 0.8.13 release blog post](https://blog.soliditylang.org/2022/03/16/solidity-0.8.13-release-announcement/)

Regards.


src/ast/ast_node_factory.ts
```diff
@@ -1143,6 +1147,7 @@ export class ASTNodeFactory {
 
         if (
             node instanceof Identifier ||
+            node instanceof IdentifierPath ||
```
cd1m0, line 1150: Good catch!

src/ast/implementation/meta/using_for_directive.ts
```diff
@@ -5,31 +5,54 @@ import { IdentifierPath } from "./identifier_path";
...
+     * One of vLibraryName or vFunctionList should always be set.
      */
-    vLibraryName: UserDefinedTypeName | IdentifierPath;
+    vLibraryName?: UserDefinedTypeName | IdentifierPath;
+
+    /**
+     * Function list for file-level using-for directives.
+     * One of vLibraryName or vFunctionList should always be set.
+     */
+    vFunctionList?: IdentifierPath[];
```
cd1m0, line 17: So the `functionList` property in the raw AST is *always* an `IdentifierPath` ? Even when single identifier are passed in?

src/compile/inference/file_level_definitions.pegjs
```diff
@@ -1,39 +1,41 @@
+{
+    // Dummy uses to silence unused function TSC errors in auto-generated code
+    expected;
```
cd1m0, line 3: Neat!

src/compile/inference/file_level_definitions.pegjs
```diff
@@ -1,39 +1,41 @@
...
+    error;
+}
+
 SourceUnit =
     __ flds: (t: FileLevelDefinition __  { return t; })* {
-        // Dummy uses to silence unused function TSC errors in auto-generated code
-        expected;
-        error;
-
-        return flds as FileLevelNode<any>[];
```
cd1m0, line 7: Could you restore the `as <type name>` annotations on the `return`s here and on other lines in this file?

I added those, so that if we forget a field in the grammar then the tsc compiler will complain when trying to compile the generated parser

cd1m0 (changes requested):
Awesome work! Just one nit in the file level parser - otherwise ready to go.

src/ast/implementation/meta/using_for_directive.ts
cd1m0: Ok, verified that its always IdentifierPath.

src/compile/inference/file_level_definitions.pegjs
blitz-1306: Will do.

blitz-1306: Addressed in https://github.com/ConsenSys/solc-typed-ast/pull/107/commits/c2ffabe33f1aca15df4949f269518b4b562049fa. Note that `UsingForDirective` have a typed variable, so it is protected from mentioned effect. Still, it uses a bit different notation.

cd1m0 (approved):

PR #109, opened 2022-03-25
Return remappings inferred by the `LocalNpmResolver` in the `CompileResult`

cd1m0:
The `compileSol` and `compileSourceString()` use the `LocalNpmResolver` class to infer any missing remappings to packages located under a node_modules directory in a parent directory of the current compiled artifact.

It would be useful to callers (especially scribble) if those entry points also reported in their `CompileResult` any additional rempapings that they inferred, on top of the remappings passed in by the user.

### DIFF
--- a/src/compile/import_resolver.ts
+++ b/src/compile/import_resolver.ts
@@ -18,16 +18,11 @@ export type Remapping = [string, string, string];
 
 export class LocalNpmResolver implements ImportResolver {
     baseDir?: string;
-    /// Array of all remappings that were "inferred" by the npm resolver
-    private _inferedRemappings: Remapping[];
+    inferedRemappings?: Map<string, Remapping>;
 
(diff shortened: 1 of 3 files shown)


blitz-1306 (approved):
I added some changes:
- First of all, I guess it worth to use a map from `fileName` to produced `Remapping`.
- It is also better to give outside control of this parameter, so I added an argument for constructor of `LocalNpmResolver` to allow setting it.
- I also added at least one test that ensures that produced map contains expected entries.

@cd1m0 Should we consider something else? WDYT?

cd1m0 (commented):
LGTM!

cd1m0:
I think you're right that there could be duplicates, so its a good decision to turn it into a map from files to their remappings. Its also strictly more information. Looks good to me as it is!

Issue #111, opened 2022-03-30
`resolveAny()` function doesn't correctly handle visiblity


cd1m0:
## Environment

Any version

## Scenario

Try calling `resolveAny('x', <FunctionDefinition for foo> ...)` on the below sample:

```
contract Base {
    uint private x;
}

contract Child is Base {
     function foo(uint x) public {}
}
```

## Expected behavior

We would expect that `resolveAny()` returns an empty set, since `x` is private in `Base` and thus not visible in `Child`

## Current behavior

`resolveAny()` will return a set containing the `VariableDeclaration` of `x`.

## Suggested solution [optional]

We should make `resolveAny` not return private variables and functions of base contracts by default. However in the future we may have use cases in [Scribble](https://github.com/ConsenSys/scribble) where we want to resolve even private fields (see discussion in https://github.com/ConsenSys/scribble/issues/163), so for now lets add a new field to `resolveAny()` `ignoreVisiblity` which is set to `false` by default, and when `ignoreVsibility` is `false`, `resolveAny` should correctly handle `private` state.

PR #112, opened 2022-03-30
Fix for #111

cd1m0:
As discussed in #111 this PR:

1. Adds a new `ignoreVisiblity` argument to `resolveAny` which is `false` by default.
2. Change `resolveAny` and related functions to not return private state vars/methods of base contracts when `ignoreVisiblity` is set to `false`.
3. Add relevant unit tests


blitz-1306 (approved):
Looks good to me.

PR #115, opened 2022-04-11
Detection of dangling documentation

### LINKED ISSUE #114: Add support for docstrings at the end of a block
We have user confusion in the Scribble tool, due to `#assert`s not appearing in the AST. Specifically on samples like these:

```solidity
for (uint i = 0; i < numbers.length; i++) {
   /// #let oldNum := numbers[i];
   numbers[i]++;
   /// #assert numbers[i] === oldNum + 1;
}
```

The `#assert` annotation is silently ignored because its docstring doesn't even appear in the AST.

The fundamental problem is that in Solidity' AST discipline each dcostring must be attached to an AST node. We should communicate with them and see if they can fix it upstream, but in the meantime perhaps we can apply the following fix when building our AST:

1. For each basic block B, let its source range be from B_start to B_end
2. Let the last statement in B be S, and let S have a source range S_start to S_end
3. **IF** we have source code, then extract the source range [S_end, B_end] from the source code
4. Try and run the existing docstring parser over the extracted source range
5. Build `StructuredDocumentation` nodes for any docstrings that were parsed in the previous step
6. Add a new field `tailDocStrings: StructuredDocumentation[]` to the `Block` nodes, that contains the parsed docstrings.

blitz-1306:
## Tasks
This PR resolves #114

## Changes
- [x] Support of detecting dangling documentation (at the end of `Block`, `UncheckedBlock`, `ContractDefinition`).
- [x] Fixes to statement nodes to properly reflect `StructuredDocumentation` as part of their children.
- [x] Related tweaks to documentation reconstruction logic for `ASTReader` postprocessor.
- [x] Related tweaks to ast-to-source writing components.
- [x] Related tweaks to affected test snapshots.

Regards.

### DIFF
--- a/src/ast/documentation.ts
+++ b/src/ast/documentation.ts
@@ -3,11 +3,21 @@ import { StructuredDocumentation } from "./implementation/meta";
 
 export interface WithPrecedingDocs {
     documentation?: string | StructuredDocumentation;
+
+    /**
+     * This field is used as a storage field for string,
+     * if string is set as value for `documentation`.
+     */
     docString?: string;
 }
 
 export interface WithDanglingDocs {
     danglingDocumentation?: string | StructuredDocumentation;
+
+    /**
+     * This field is used as a storage field for string,
+     * if string is set as value for `danglingDocumentation`.
+     */
     danglingDocString?: string;
 }
 
@@ -27,6 +37,10 @@ export function getDocumentation(
         if (child instanceof StructuredDocumentation) {
             const childLoc = child.sourceInfo;
 
+            /**
+             * Note that preceding documentation nodes are
+             * EXCLUDED from source range of parent node.
+             */
             if (childLoc.offset <= ownLoc.offset) {
                 return child;
             }
@@ -77,6 +91,10 @@ export function getDanglingDocumentation(
         if (child instanceof StructuredDocumentation) {
             const childLoc = child.sourceInfo;
 
+            /**
+             * Note that preceding documentation nodes are
+             * INCLUDED from source range of parent node.
+             */
             if (childLoc.offset > ownLoc.offset) {
                 return child;
             }
--- a/src/ast/documentation.ts
+++ b/src/ast/documentation.ts
@@ -1,7 +1,7 @@
 import { ASTNode, ASTNodeWithChildren } from "./ast_node";
 import { StructuredDocumentation } from "./implementation/meta";
 
-export interface WithPreceedingDocs {
+export interface WithPrecedingDocs {
     documentation?: string | StructuredDocumentation;
     docString?: string;
 }
@@ -12,7 +12,7 @@ export interface WithDanglingDocs {
 }
 
 export function getDocumentation(
-    node: WithPreceedingDocs & ASTNodeWithChildren<ASTNode>
+    node: WithPrecedingDocs & ASTNodeWithChildren<ASTNode>
 ): string | StructuredDocumentation | undefined {
     if (node.docString !== undefined) {
         return node.docString;
@@ -37,7 +37,7 @@ export function getDocumentation(
 }
 
 export function setDocumentation(
-    node: WithPreceedingDocs & ASTNodeWithChildren<ASTNode>,
+    node: WithPrecedingDocs & ASTNodeWithChildren<ASTNode>,
     value: string | StructuredDocumentation | undefined
 ): void {
     const old = node.documentation;
--- a/src/ast/documentation.ts
+++ b/src/ast/documentation.ts
@@ -1,0 +1,112 @@
+import { ASTNode, ASTNodeWithChildren } from "./ast_node";
+import { StructuredDocumentation } from "./implementation/meta";
+
+export interface WithPreceedingDocs {
+    documentation?: string | StructuredDocumentation;
+    docString?: string;
+}
+
+export interface WithDanglingDocs {
+    danglingDocumentation?: string | StructuredDocumentation;
+    danglingDocString?: string;
+}
+
+export function getDocumentation(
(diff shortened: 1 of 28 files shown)


src/ast/documentation.ts
```diff
@@ -0,0 +1,112 @@
+import { ASTNode, ASTNodeWithChildren } from "./ast_node";
+import { StructuredDocumentation } from "./implementation/meta";
+
+export interface WithPrecedingDocs {
+    documentation?: string | StructuredDocumentation;
+    docString?: string;
```
cd1m0, line 6: Why do we have 2 fields - `docString` and `documentation`? It feels like having a single `documentation` field should be enough

src/ast/postprocessing/structured_documentation_reconstruction.ts
```diff
@@ -146,19 +196,47 @@ export class StructuredDocumentationReconstructingPostprocessor
...
-        if (structDocNode === undefined) {
-            return;
+            if (preceding) {
+                preceding.id = context.lastId + 1;
+
+                context.register(preceding);
+
+                node.documentation = preceding;
+
+                if (preceding.parent !== node) {
```
cd1m0, line 214: Is this `if` neccessary? Seems like you can always just do `preceding.parent = node;`

src/ast/postprocessing/structured_documentation_reconstruction.ts
```diff
@@ -146,19 +196,47 @@ export class StructuredDocumentationReconstructingPostprocessor
...
+            if (dangling) {
+                dangling.id = context.lastId + 1;
 
-        node.documentation = structDocNode;
+                context.register(dangling);
 
-        structDocNode.parent = node;
+                node.danglingDocumentation = dangling;
+
+                if (dangling.parent !== node) {
```
cd1m0, line 235: Same question about the if

src/ast/postprocessing/structured_documentation_reconstruction.ts
```diff
@@ -81,12 +114,22 @@ export class StructuredDocumentationReconstructor {
 
         let stopOnNextGap = false;
 
+        const rxRemoveGarbage = /^[^/]+/;
```
cd1m0, line 117: Is this rx meant to remove leading spaces before the `///` or `*` ?

cd1m0 (changes requested):
Just a few questions to better understand the PR

src/ast/documentation.ts
blitz-1306: Well, `docString` is defined specifically to store a `string`, when it is set as a value for the property. The `documentation` is the external property that is implemented by **getter**, that works in following order:
- If `docString` **is not** `undefined`, then retuns `docString`.
- If `docString` **is** `undefined`, then check `children` from very beginning and return `StructuredDocumentation` node that is preceeding current one: struct docs source range is out of the parent node source range (in other words they are excluded from parent's source range by compiler).
- If there is none of above applicable - return `undefined`.

Same applies to the `danglingDocumentation` with few exceptions:
- The `children` are checked from very end.
- Then `StructuredDocumentation` node need to be included into parent's source range.

src/ast/postprocessing/structured_documentation_reconstruction.ts
blitz-1306: Will address

src/ast/postprocessing/structured_documentation_reconstruction.ts
blitz-1306: Will address

src/ast/postprocessing/structured_documentation_reconstruction.ts
blitz-1306: Not only the spaces, but everything before the first `/`. This is because the cases with dangling nodes:
```solidity
{
    /// dangling
}
```
The AST does not specifically give any information about where the internal whitespace starts, so we **have to** look from very begining of `Block`'s source range, that includes `{\n\s` before `/`. This is even more uglier when we have ```solidity
unchecked {
    /** dangling */
}
```
or
```solidity
contract {
    /// dangling
}
```
Due to this, we have to remove any possible garbage before first leading `/` and then check if it is `///` or `/**`.

blitz-1306: I will document it better.

blitz-1306: Addressed in https://github.com/ConsenSys/solc-typed-ast/pull/115/commits/872be86fefb3d356d2552929b0a260684ffd6801

src/ast/postprocessing/structured_documentation_reconstruction.ts
blitz-1306: Addressed in https://github.com/ConsenSys/solc-typed-ast/pull/115/commits/872be86fefb3d356d2552929b0a260684ffd6801

src/ast/postprocessing/structured_documentation_reconstruction.ts
blitz-1306: Addressed in https://github.com/ConsenSys/solc-typed-ast/pull/115/commits/872be86fefb3d356d2552929b0a260684ffd6801

src/ast/documentation.ts
blitz-1306: Added a bit more documentation to `docString` and `danglingDocString` in https://github.com/ConsenSys/solc-typed-ast/pull/115/commits/872be86fefb3d356d2552929b0a260684ffd6801

cd1m0 (approved):
LGTM

PR #118, opened 2022-05-04
Support `--base-path` and `--include-path`s

blitz-1306:
## Changes
- [x] Updated dependencies.
- [x] Dropped usage of [`minimist`](https://www.npmjs.com/package/minimist) in favor of [`Commanter.js`](https://www.npmjs.com/package/commander) due to correct array arguments support. Also, extremely simplified flow of executable (`sol-ast-compile` utility).
- [x] Function `compileSol()` and `compileSourceString()` now can receive `PathOptions` object, that allows to configure path remappings, base path (root of the contracts setup) and include paths (optional paths to look for dependency contracts).
- [x] FileSystem import resolver can now accept "base" and "include" paths. It also checks if input file name is relative to them.
- [x] Introduced CLI options `--base-path` and `--include-path`.
- [x] Removed `makeExpression()`, `makeStatement()`, `makeTypeName()` and `makeNode()` from `ASTNodeFactory` as do not seem to have further use.
- [x] Fixed affected tests and snapshots.

## Notes
- Changes are affecting external API, so this is a breaking change that would require major version bump.
- Unsure if FileSystem import resolver should produce `inferedRemappings`. If so, unsure how to implement their support.
- Order of produced source units has changed and that is why so much snapshots were affected.
- I took some inspiration from SolcJS, particularly:
  - https://github.com/ethereum/solc-js/blob/99eafc2ad13d5bfba7886de659a6638814d95078/solc.ts#L26-L54 
  - https://github.com/ethereum/solc-js/blob/99eafc2ad13d5bfba7886de659a6638814d95078/solc.ts#L62-L78
  - https://github.com/ethereum/solc-js/blob/99eafc2ad13d5bfba7886de659a6638814d95078/solc.ts#L89-L113

Regards.


test/integration/sourcemap/snapshot.spec.ts
```diff
@@ -79,7 +79,7 @@ describe("Source map snapshot tests", () => {
                     const result = parts.join("\n") + "\n";
 
                     // Uncomment next line to update snapshots
-                    // fse.writeFileSync(sample, result, { encoding: "utf-8" });
+                    fse.writeFileSync(sample, result, { encoding: "utf-8" });
```
cd1m0, line 82: This should be commented right?

cd1m0 (changes requested):
Awesome work! Just 2 questions (some stuff I don't quite understand)  and 1 nit

test/integration/sourcemap/snapshot.spec.ts
blitz-1306: Right, will fix.

src/compile/import_resolver.ts
```diff
@@ -8,27 +8,55 @@ export interface ImportResolver {
...
+            this.includePaths ? this.includePaths : []
+        );
+
+        for (const prefix of prefixes) {
+            let candidate: string;
+
+            if (prefix) {
+                const relative = path.relative(prefix, fileName);
+
+                candidate = path.join(prefix, relative.startsWith("../") ? fileName : relative);
```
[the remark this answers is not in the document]
blitz-1306: We try to figure out if cimpiler input source file path is relative to base path. In that case we want to avoid double prefix of base path. Relative path would contain `../` in case if they are not relative. Consider looking [here](https://github.com/ethereum/solc-js/blob/99eafc2ad13d5bfba7886de659a6638814d95078/solc.ts#L108).

test/integration/sourcemap/snapshot.spec.ts
blitz-1306: Addressed in https://github.com/ConsenSys/solc-typed-ast/pull/118/commits/71fe51fc8f15cb2fb7dedb1e7f639465088f3425

src/compile/utils.ts
```diff
@@ -238,20 +249,30 @@ export async function compileSol(
...
     const parsedRemapping = parsePathRemapping(remapping);
+
     const files = new Map<string, string>();
     const visited = new Set<string>();
 
+    const isDynamicBasePath = pathOptions.basePath === undefined;
+
     for (const fileName of fileNames) {
-        const sourceCode = await fse.readFile(fileName, "utf-8");
+        const resolvedFileName = fsResolver.resolve(fileName) || path.resolve(fileName);
```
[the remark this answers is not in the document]
blitz-1306: It _was_ a hack due to FileSystem resolver was not always detecting file path for some edge cases. Addressed in https://github.com/ConsenSys/solc-typed-ast/pull/118/commits/71fe51fc8f15cb2fb7dedb1e7f639465088f3425

cd1m0 (approved):

PR #119, opened 2022-05-20
Report resolved paths in CompileResult

cd1m0:
The compilation helper functions perform several complex tasks to resolve imported files:

- do path remappings;
- look for files under `node_modules` and infer some remappings;
- take into account `--base-path` and `--include-paths` when looking for files.

As a result, even though in the compilation result there is some file X/Y/Z.sol, its not obvious to the callers of `compile*()` where that file may live on disk. This is especially useful to Scribble, which needs to move files around when arming.

To support this, this PR suggest to add a new field to `CompileResult` - `resolvedFileNames`:

```
    /**
     * Map from file names appearing in the files map, to the
     * actual resolved paths on disk (if any). For compileJSONData this
     * map is empty (since nothing was resolved on disk)
     */
    resolvedFileNames: Map<string, string>;
```

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6,7 +6,7 @@
     "packages": {
         "": {
             "name": "solc-typed-ast",
-            "version": "9.1.3",
+            "version": "10.0.0",
             "license": "Apache-2.0",
             "dependencies": {
                 "axios": "^0.27.2",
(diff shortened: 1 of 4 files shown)


blitz-1306 (approved):
Great improvement, thanks!

PR #130, opened 2022-07-12
Move to GitHub Actions instead of Drone.io

blitz-1306:


blitz-1306:
@cd1m0 Hello. Could you, please, disable Drone hooks in admin section of this repository? After migrating to GitHub Actions, we would not need that anymore. Thanks.

cd1m0 (approved):
LGTM

Issue #132, opened 2022-07-15
Undefined `kind` field on fallback `FunctionDefinition` node on 0.4.24


cd1m0:
## Environment

This happens with the latest solc-typed-ast (10.1.3) when compiling contracts written for  solidity 0.4.24.

## Scenario

Try and run `sol-ast-compile` on the follwoing contract:

```
pragma solidity 0.4.24;

contract FBNoArgs {
    int8 public lastCall = 0;
    
    function() external payable {
        lastCall = 1;
    }
    
    function id(uint x) public returns (uint) {
        lastCall = 2;
        return x;
    }
}
```

You will notice that the `kind` field of the `FunctionDefinition` node corresponding to the fallback function is undefined:

```
        FunctionDefinition #12
            id: 12
            src: "84:57:0"
            implemented: true
            virtual: false
            scope: 27
            kind: undefined
            name: ""
            visibility: "external"
            stateMutability: "payable"
...
```
## Expected behavior

The field should be "fallback".

## Current behavior

The field is undefined.

## Description [optional]

I suspect the issue is that on 0.4.x that field is undefined for fallback methods. The legacy processor for function definition nodes has [logic](https://github.com/ConsenSys/solc-typed-ast/blob/master/src/ast/legacy/function_definition_processor.ts#L68) for detecting the kind field. However the modern one doesn't and just takes the value of kind [straight](https://github.com/ConsenSys/solc-typed-ast/blob/master/src/ast/modern/function_definition_processor.ts#L20) from the raw ast.

## Suggested solution [optional]

Lets use the same detection logic for both processors?

blitz-1306:
This bug is fixed in #133 and will be fully addressed in next package release.

PR #133, opened 2022-07-15
Fix for #132

cd1m0:
What title says - fix for #132. Re-use the detection logic of the legacy processor in the modern one.


blitz-1306 (approved):
I moved the shared logic to utils file and added test sample. Looks good now.

PR #138, opened 2022-07-28
Tweak `eq()` to do an early return if `a` and `b` are strictly equal

blitz-1306:
## Changes
- [x] Updated peggy-related dependency after upstream fixes (https://github.com/metadevpro/ts-pegjs/pull/80).
- [x] Tweaked `eq()` to do an early return if `a` and `b` are strictly equal.

Regards.


cd1m0 (approved):
LGTM

PR #142, opened 2022-08-05
Alternative type inference

cd1m0:
1. Problem

Currently we rely on the typeString parser to get the type of an Expression. The typeString parser is fragile (breaks for some int/string literals for example) and requires us to do a bunch of fix-ups (for example wrap conrtact types with pointers). 

2. Solution

Add a new `InferType` class that provides a `typeOf(Expression)` function that computes the type of a given expression. Supporting this properly required A LOT of fixes. Specifically we had to:

- Add a bunch of new types to support various ASTNodes
- Add a limited form of polymorphism to cleanly support some polymorphic builtins
- Add the inference code
- Add a simple constant expression evaluator to infer precise types for constant expressions
- Add a bunch of tests comparing the results of `InferType` to the old typeString parser (see infer.spec.ts)
- etc...

3. Remaining work

There are several pieces of work remaining here:

- [ ]  The constant expression evaluation code is incomplete (e.g. all the bitwise operations are not implemented)
- [ ] The constant expression evaluation code needs tests
- [ ] The constant expression evaluation code uses `decimal.js`. I am not sure this is the right decision. @blitz-1306 what do you think?
- [ ] Fix any TODOs remaining in this PR

4. Note:

- This is a VERY INTERFACE BREAKING CHANGE. The behavior of specializeType and the type string parser has changed subtly. Also we are adding a whole new interface, and should begin considering the typestring parser as deprecated. So this requires a major release.
- I had to slightly weaken the typeString correctness tests  (typestring.spec.ts) due to subtle changes in the TypeNode printing. On the one hand, this is ok as we will soon deprecate the typeString parser. On the other - we will likely still use the typeString parser for correctness testing of the actual type inference engine. So not sure yet if this was the right move.

PR #159, opened 2022-10-18
Rework deprecated functionality

blitz-1306:
## Preface
Previously we introduced new `InferType` component in #142 and #152 PRs, also in [10.3.0 release](https://github.com/ConsenSys/solc-typed-ast/releases/tag/v10.3.0). We also realized that it would impact on backward compatibility due to older type inference functions would significantly change their interface. We postponed this step for a while to inform dependent packages about further changes. This PR reworks a lot of deprectaed functions and adapts them to a newer type inference approach.

## Changes
- [x] Rework deprecated functionality to fully support newer type inference component.
- [x] Use typeString parser with `ModuleType` and `InaccessibleDynamicType` for tests only (remove from common codebase).
- [x] Updated dependencies.
- [x] Fix affected tests.

## Notes
- This PR causes BC break and would require to bump major version of the package during release.
- Created a followup issue #160.
- We test this PR agains one of the consumers here: https://github.com/ConsenSys/scribble/pull/200.

Regards.


blitz-1306:
I think we should not merge until we stabilize Scribble behavior, while it is using this branch. See failures in https://github.com/ConsenSys/scribble/pull/200.

src/types/builtins.ts
```diff
@@ -41,8 +40,7 @@ export const typeContract = new BuiltinFunctionType(
             new Map([
                 ["name", [[types.stringMemory, ">=0.5.5"]]],
                 ["creationCode", [[types.bytesMemory, ">=0.5.3"]]],
-                ["runtimeCode", [[types.bytesMemory, ">=0.5.3"]]],
-                ["interfaceId", [[types.bytes4, ">=0.6.7"]]]
+                ["runtimeCode", [[types.bytesMemory, ">=0.5.3"]]]
```
cd1m0, line 43: Why did we remove `interfaceId` as a builtin?

src/types/utils.ts
```diff
@@ -570,13 +508,11 @@ export function castable(fromT: TypeNode, toT: TypeNode, compilerVersion: string
...
         if (toT instanceof AddressType) {
-            // We can implicitly cast from contract to payable address if it has a payable recieve/fallback function
+            // We can implicitly cast from contract to payable address if it has a payable receive/fallback function
             if (toT.payable) {
-                const fns = getFallbackRecvFuns(fromT.definition);
-
-                return (
-                    fns.length > 0 &&
-                    forAll(fns, (fn) => fn.stateMutability === FunctionStateMutability.Payable)
+                return forAny(
```
cd1m0, line 513: Good catch on the forAny

cd1m0 (changes requested):
1 note and 1 question.

src/ast/definitions.ts
```diff
@@ -239,22 +241,16 @@ function* lookupInContractDefinition(
...
-                    sigHash = child.canonicalSignatureHash(ABIEncoderVersion.V2);
-                }
+                /**
+                 * Its a safe to assume V2 as its backward-compatible and we only use it internally here
+                 */
+                const sigHash =
+                    child instanceof FunctionDefinition ||
+                    child instanceof EventDefinition ||
+                    (child instanceof VariableDeclaration &&
+                        child.visibility === StateVariableVisibility.Public)
```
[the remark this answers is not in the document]
cd1m0: @blitz-1306 feel free to mark this reviewed as soon as you take a look at the issue :)

src/types/builtins.ts
cd1m0: Got it!

src/types/infer.ts
```diff
@@ -1390,15 +1300,15 @@ export class InferType {
...
             // For solidity <0.5.0 contract variables are implicitly castable to address
             if (lt(this.version, "0.5.0")) {
                 builtinT = addressBuiltins.getFieldForVersion(node.memberName, this.version);
             }
 
             if (fieldT) {
-                if (this.isFunctionCallExternal(node, baseT)) {
-                    fieldT = this.changeLocToMemory(fieldT);
-                }
+                fieldT = this.changeLocToMemory(fieldT);
```
cd1m0, line 1311: Why is this no longer in an if?

src/types/infer.ts
```diff
@@ -2092,4 +2007,413 @@ export class InferType {
...
+            for (const member of retType.to.definition.vMembers) {
+                const memberT = member.vType;
+
+                assert(
+                    memberT !== undefined,
+                    "Unexpected untyped struct member",
+                    retType.to.definition
+                );
+
+                if (memberT instanceof Mapping || memberT instanceof ArrayTypeName) {
```
cd1m0, line 2075: I wonder what should we do if `memberT` is a struct?

cd1m0 (changes requested):
One last question, sorry I forgot one file last review

src/types/infer.ts
cd1m0: We don't have to fix this right now, cause it is not making the behavior worse. Just wondering what the correct semantic is. Will make a note of this in an issue and add it here.

cd1m0: Tracked under https://github.com/ConsenSys/solc-typed-ast/issues/169

src/types/infer.ts
blitz-1306: Well, previous logic actually was working that way... Let me explain:

First of all, this if is nested in
```typescript
if (baseT instanceof UserDefinedType && baseT.definition instanceof ContractDefinition) {
```
So, `baseT` is UserDefinedType, that is a contract...Then we call
```typescript
if (fieldT) {
    if (this.isFunctionCallExternal(node, baseT)) {
```
Where `node` is a `MemberAccess` and `baseT` is UDT of contract. The `isFunctionCallExternal()` back at the time had following implementation:
```typescript
private isFunctionCallExternal(callee: ASTNode, calleeT: TypeNode): boolean {
    if (!(callee instanceof MemberAccess)) { // <- This would never be true
        return false;
    }

    const name = callee.memberName;

    if (
        calleeT instanceof BuiltinFunctionType &&
        ["call", "callcode", "staticcall", "delegatecall", "transfer", "send"].includes(name)
    ) {
        return true;
    }

    const baseT = this.typeOf(callee.vExpression);
    // this is the same thing as we checked in first IF, as initial fieldT was computed in the same way

    return baseT instanceof UserDefinedType && baseT.definition instanceof ContractDefinition;
    // this would always be true
}
```
So, this check seemed to be obsolete. Contract archive run also didn't show any difference.

src/types/utils.ts
```diff
@@ -429,6 +348,25 @@ export function getABIEncoderVersion(
...
+    }
+
+    if (
+        call.vFunctionCallType === ExternalReferenceType.Builtin &&
+        CALL_BUILTINS.includes(call.vFunctionName)
+    ) {
+        return true;
+    }
+
+    if (call.vExpression.typeString.includes(FunctionVisibility.External)) {
```
cd1m0, line 363: Can the type string be something like this `function (externalStructName memory) internal`? This might be fragile. Lets file an issue for a cleaner implementation of `isFunctionCallExternal`

cd1m0 (approved):
lgtm. Some issues for follow up work.

src/types/utils.ts
blitz-1306: This is tracked by #171.

PR #174, opened 2022-11-23
Expose constant expression evaluator logic and tweak `getTypeForCompilerVersion()`

### LINKED ISSUE #170: Nit: Change signature of getTypeForCompilerVersion
In Scribble the signature of `getTypeForCompilerVersion` is:

```typescript
function getTypeForCompilerVersion(
    typing: TypeNode | [TypeNode, string],
    compilerVersion: string
): TypeNode | undefined {
```

while in solc-typed-ast its:

```typescript
export function getTypeForCompilerVersion(
    typing: [TypeNode, string] ,
    compilerVersion: string
): TypeNode | undefined {
```

Change the solc-typed-ast version to look like the scribble version and delete the scribble version.

### LINKED ISSUE #173: Nit: Expose `evalLiteral`, `evalUnary`, `evalBinary*` to scribble
In order to replace `isImplictlyCastable` in scribble with `castable` from solc-typed-ast, we first need to add the literal values to the `IntLiteralType`s, that we compute in `tc()`. However this then breaks `unifyTypes`, as two different literals are not *castable* one to another. In order to get around this problem, we need to implement constant expression evaluation in scribble, similarly to solc-typed-ast, so we can compute an exact `IntLiteralType` for constant expressions. Since Scribble's arithmetic follows the same semantics as Solidity, its silly to duplicate the code we already have in solc-typed-ast. Instead lets expose it so that scribble can use it. Namely we should expose the following functions:

`function evalLiteralImpl(kind: LiteralKind, val: Value): Value {`
`function evalUnaryImpl(op: string, value: Value): Value {`
`function evalBinaryImpl(op: string, lVal: Value rVal: Value): Value {`

blitz-1306:
## Tasks
This PR fixes #170 and fixes #173.

## Changes
- Allow `TypeNode` as a `typing` for `getTypeForCompilerVersion()`.
- Introduce `evalLiteralImpl()`, `evalUnaryImpl()`, `evalBinaryImpl()`.
- Export constant expression evaluator internal functions.

## Notes
This PR slightly changes `EvalError` class.

Regards.

### DIFF
--- a/src/types/eval_const.ts
+++ b/src/types/eval_const.ts
@@ -176,7 +176,7 @@ export function evalUnaryImpl(operator: string, value: Value): Value {
             return ~value;
         }
 
-        throw new EvalError(`Expected ${str(value)} to be an bigint`);
+        throw new EvalError(`Expected ${str(value)} to be a bigint`);
     }
 
     if (operator === "+") {
@@ -184,7 +184,7 @@ export function evalUnaryImpl(operator: string, value: Value): Value {
             return value;
         }
 
-        throw new EvalError(`Expected ${str(value)} to be an bigint or decimal`);
+        throw new EvalError(`Expected ${str(value)} to be a bigint or a decimal`);
     }
 
     if (operator === "-") {
@@ -196,7 +196,7 @@ export function evalUnaryImpl(operator: string, value: Value): Value {
             return -value;
         }
 
-        throw new EvalError(`Expected ${str(value)} to be an bigint or decimal`);
+        throw new EvalError(`Expected ${str(value)} to be a bigint or a decimal`);
     }
 
     throw new EvalError(`Unable to process ${operator}${str(value)}`);
--- a/src/types/eval_const.ts
+++ b/src/types/eval_const.ts
@@ -204,7 +204,7 @@ export function evalUnaryImpl(operator: string, value: Value): Value {
 
 export function evalBinaryImpl(operator: string, left: Value, right: Value): Value {
     if (BINARY_OPERATOR_GROUPS.Logical.includes(operator)) {
-        if (!(typeof left === "boolean" && typeof left === "boolean")) {
+        if (!(typeof left === "boolean" && typeof right === "boolean")) {
             throw new EvalError(`${operator} expects booleans not ${str(left)} and ${str(right)}`);
         }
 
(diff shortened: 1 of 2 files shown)


src/types/eval_const.ts
```diff
@@ -149,212 +159,213 @@ function evalLiteral(expr: Literal): Value {
...
-
-    if (op === "||") {
-        return lVal || rVal;
-    }
-
-    throw new Error(`Unknown logic op ${op}`);
-}
+export function evalBinaryImpl(operator: string, left: Value, right: Value): Value {
+    if (BINARY_OPERATOR_GROUPS.Logical.includes(operator)) {
+        if (!(typeof left === "boolean" && typeof left === "boolean")) {
```
cd1m0, line 207: I think the second typeof should be `typeof right` no?

src/types/eval_const.ts
```diff
@@ -364,51 +375,51 @@ function evalBinary(expr: BinaryOperation): Value {
...
 
-    if (expr instanceof BinaryOperation) {
-        return evalBinary(expr);
+    if (node instanceof BinaryOperation) {
+        return evalBinary(node);
     }
 
-    if (expr instanceof TupleExpression) {
-        return evalConstantExpr(expr.vOriginalComponents[0] as Expression);
+    if (node instanceof TupleExpression) {
```
cd1m0, line 395: I think the condition should be `node instanceof TupleExpression && node.vOriginalComponents.length === 1`

src/types/eval_const.ts
```diff
@@ -364,51 +375,51 @@ function evalBinary(expr: BinaryOperation): Value {
...
+    if (node instanceof Identifier) {
+        const decl = node.vReferencedDeclaration;
 
         if (decl instanceof VariableDeclaration) {
             return evalConstantExpr(decl.vValue as Expression);
         }
     }
 
-    if (expr instanceof FunctionCall) {
+    if (node instanceof FunctionCall) {
```
cd1m0, line 413: Lets be precise here and check also that `node.kind === FunctionCallKind.typeConversion`

cd1m0 (changes requested):
A few nits

src/types/eval_const.ts
cd1m0: The call to `isConstant` does that check.

src/types/eval_const.ts
blitz-1306: Fixed in https://github.com/ConsenSys/solc-typed-ast/pull/174/commits/37f2e40caa13d606c7e3a392b09e53f218c889e9

cd1m0 (approved):

PR #175, opened 2022-11-28
More influence of `ABIEncoder`

### LINKED ISSUE #167: Fix getABIEncoderVersion function
Currently the signature of the `getABIEncoderVersion` function computes one version for multiple source units:

```
export function getABIEncoderVersion(
    units: SourceUnit[],
    compilerVersion: string
): ABIEncoderVersion {
```

However this is incorrect. You can have in a **single** compilation, multiple source units with different `ABIEncoderVersion`s. To fix this we should:

1. Change the signature of `getABIEncoderVersion` to:

```
export function getABIEncoderVersion(
    units: SourceUnit,
    compilerVersion: string
): ABIEncoderVersion {
```

to force consumers of solc-typed-ast to deal with each unit separately.

2. Fix `getABIEncoderVersion` to handle the case when the `abicoder` pragma is in an imported unit

3. Add tests with:

3.1 One source unit with 2 conflicting encoder pragmas

3.2 One source unit with 2 non-conflicting (same) encoder pragmas

3.3 Source unit A, without a pragma, that imports unit B, with a pragma

3.4 Source unit A, with a pragma, that imports unit B, with a conflicting pragma

### LINKED ISSUE #168: resolveAny() needs ABIEncoderVersion as well
We have currently hardcoded `ABIEncoderVersion.V2` inside `lookupInContractDefinition`. However this is potentially not compatible with older code. So we should find a way to pass in the abi version here.

 At first I thought we should just add abicoder version as a field to InferType.

However in a single compilation, you can have multiple SourceUnits with the same compiler version, but different ABI versions. So it doesn't seem like a good idea to keep it in InferType, since we were planning to have one InferType object per compilation.

So for now, as unpleasant as it is, perhaps we need to pass these manually?

Alternatively, we can try and add a computed getter on SourceUnit nodes that specifies their abi version. Then for any other node we can get the corresponding abi coder version by doing node.getClosestParentByType(SourceUnit).abiVersion. Not sure, we should discuss more. Probably file an issue.

Needs discussion

### LINKED ISSUE #169: `InferType.getterArgsAndReturns` doesn't handle nested structs
The logic in `InferType.getterArgsAndReturns` doesn't correctly handle nested structs. For samples like this:

```
contract Foo {
    struct S1 {
        uint x;
        uint y;
    }

    struct S2 {
        bool b;
        S1 s;
        address a;
    }

    S2 public s;
}
```

It should compute a return type for `s` like this `(bool, (uint, uint), address)`. Instead it will compute something like `(bool, S1 memory, address)`. Fix this.

blitz-1306:
## Tasks
This PR fixes #167, fixes #168, fixes #169.

## Preface
We had ABI encoder version switches here and there, while there was no general type system component. Now it appears that type system significantly affected by ABI encoder version - getter signatures are composed differently, overloading resolution is also affected by this. Attemps to introduce encoder as and argument are causing insane set of interface breaking changes, however we can reduce impact by adding it as a state parameter of `InferType`.

## Changes
- [x] Introduce ABI encoder version as an `InferType` field.
- [x] Adjust multiple functions to respect ABI encoder, including `resolve*` family (#168).
- [x] Removed `encoderVersion` argument for `InferType` methods (it is now used from field).
- [x] Removed assertion checks for nested struct type arguments during compostion of public getter types when ABI encoder V1 is used (#169).
- [x] Tweak `getABIEncoderVersion()` to compute encoder version for single source unit only (#167).
- [x] Allow `type(<interface>)` members to be applied to `abstract` contracts.
- [x] Adjust codebase to respect changes. Fix tests.

Regards.

### DIFF
--- a/src/ast/definitions.ts
+++ b/src/ast/definitions.ts
@@ -241,9 +241,6 @@ function* lookupInContractDefinition(
                     continue;
                 }
 
-                /**
-                 * Its a safe to assume V2 as its backward-compatible and we only use it internally here
-                 */
                 const sigHash =
                     child instanceof FunctionDefinition ||
                     child instanceof EventDefinition ||
--- a/src/ast/definitions.ts
+++ b/src/ast/definitions.ts
@@ -1,7 +1,7 @@
 import { gte, lt } from "semver";
 import { FunctionVisibility } from ".";
 import { assert, forAll, pp } from "../misc";
-import { ABIEncoderVersion, InferType } from "../types";
+import { InferType } from "../types";
 import { ASTNode } from "./ast_node";
 import { StateVariableVisibility } from "./constants";
 import { ContractDefinition } from "./implementation/declaration/contract_definition";
@@ -249,7 +249,7 @@ function* lookupInContractDefinition(
                     child instanceof EventDefinition ||
                     (child instanceof VariableDeclaration &&
                         child.visibility === StateVariableVisibility.Public)
-                        ? inference.signatureHash(child, ABIEncoderVersion.V2)
+                        ? inference.signatureHash(child)
                         : undefined;
 
                 if (sigHash !== undefined) {
--- a/src/types/infer.ts
+++ b/src/types/infer.ts
@@ -193,7 +193,10 @@ function markFirstArgImplicit<T extends FunctionType | FunctionLikeSetType<Funct
 }
 
 export class InferType {
-    constructor(public readonly version: string) {}
+    constructor(
+        public readonly version: string,
+        public readonly encoderVersion: ABIEncoderVersion
+    ) {}
 
     /**
      * Infer the type of the assignment `node`. (In solidity assignments are expressions)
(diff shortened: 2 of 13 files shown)


src/ast/definitions.ts
```diff
@@ -241,15 +241,12 @@ function* lookupInContractDefinition(
...
-                /**
-                 * Its a safe to assume V2 as its backward-compatible and we only use it internally here
-                 */
                 const sigHash =
                     child instanceof FunctionDefinition ||
                     child instanceof EventDefinition ||
                     (child instanceof VariableDeclaration &&
                         child.visibility === StateVariableVisibility.Public)
-                        ? inference.signatureHash(child, ABIEncoderVersion.V2)
+                        ? inference.signatureHash(child)
```
blitz-1306, line 249: This causes slight regression in contract archive:
```
Error: Getters of struct return type are not supported by ABI encoder v1
```
Apprx 13 samples are failing.

src/types/infer.ts
```diff
@@ -878,7 +881,9 @@ export class InferType {
 
         if (innerT instanceof UserDefinedType && innerT.definition instanceof ContractDefinition) {
             const resTemplateT =
-                innerT.definition.kind === ContractKind.Interface ? typeInterface : typeContract;
+                innerT.definition.kind === ContractKind.Interface || innerT.definition.abstract
```
blitz-1306, line 884: It apperars that Solidity allow `intefaceId` to be used on `abstract` contracts.

cd1m0 (approved):
LGTM

PR #179, opened 2023-02-14
Fix inaccurate tuple type inference

cd1m0:
Solidity tuples can have empty elements. E.g:

`(1, ,"foo", , false)`

The type inference currently ignores empty elements and for the above tuple would compute a type `(int_literal 1, string memory, bool)`. However this is not accurate. This is the type of the tuple `(1, "foo", false)`. Fix this by allowing element types of the tuple type to also be `null`. 

This change simplifies decoding tuple assignments with potentially empty elements.


blitz-1306 (approved):
Looks good to me. Added few improvements in 64e968a.

PR #183, opened 2023-03-10
Handle string literals for constant expression evaluator

blitz-1306:
## Changes
- [x] Add support for string literals for equality and comparison ops in constant expression evaluator.

An example:
```solidity
contract Sample {
    function test() public {
        assert(bytes1(0x58) == 'X');
        assert(bytes2(0x5859) == 'XY');
        assert(bytes1(0x58) == hex'58');
        assert(bytes2(0x5859) == hex'5859');
        assert(bytes1(0x00) == '');
        assert(bytes1(0x00) == hex'');
    }
}
```

Regards.


src/types/eval_const.ts
```diff
@@ -220,19 +228,16 @@ export function evalBinaryImpl(operator: string, left: Value, right: Value): Val
...
             throw new EvalError(
                 `${operator} not allowed for strings ${str(left)} and ${str(right)}`
             );
         }
 
-        let isEqual: boolean;
+        const leftDec = promoteToDec(left);
+        const rightDec = promoteToDec(right);
 
-        if (left instanceof Decimal && right instanceof Decimal) {
```
cd1m0, line 231: I think this change may break constant eval of `true == true`, since bool is not promotable to Decimal.

cd1m0: Maybe change this to:

```
if (left instanceof Decimal || right instanceof Decimal) {
        const leftDec = promoteToDec(left);
        const rightDec = promoteToDec(right);
        
        isEqual = leftDec.equals(rightDec);
} else {
       isEqual = left === right;
}

cd1m0 (changes requested):
One nit

src/types/eval_const.ts
blitz-1306: Good note, will fix.

blitz-1306: Addressed slightly different way in d06e427. Consider.

cd1m0 (approved):

PR #184, opened 2023-03-15
Support quoted version pragma for compiler detection

blitz-1306:
## Changes
- [x] Add support for double- or single-qouted version pragma directive for automatic compiler detection version strategy.

An example:
```solidity
pragma solidity "0.8.17";
pragma solidity '0.8.17';

contract Test {}
```

Regards.


cd1m0 (approved):
lgtm

PR #191, opened 2023-03-27
Support `TryCatchClause` for `InferType.signature*`

### LINKED ISSUE #190: Allow `InferType.signature`/`InferType.signatureHash` to accept `TryCatchClause`
What title says. Currently we have logic in other repos for computing signatures of hashes like this:

```
        const argTypes: string[] = clause.vParameters
            ? clause.vParameters.vParameters.map((decl) =>
                  infer.signature(decl, ctx.abiEncoderVersion)
              )
            : [];
        const sig = `${clause.errorName}(${argTypes.join(",")})`;
        const hashStr = encodeFuncSignature(sig, true);
```

We should just move this inside `infer` and be able to call `infer.signature()` on a `TryCatchClause`.

If the `TryCatchClause` doesn't have an errorName, we can just return an empty signature, and 0x0 as signature hash.

blitz-1306:
## Tasks
Resolves #190.

## Changes
- [x] Support passing `TryCatchClause` to `InferType.signature()` and `InferType.signatureHash()`.
- [x] Fix camelCase for `InferType.specializeBuiltinTypeToCall()`.

## Notes
Unsure about tests. Would be great to have a related sample.

Regards.

### DIFF
--- a/src/types/infer.ts
+++ b/src/types/infer.ts
@@ -41,6 +41,7 @@ import {
     SourceUnit,
     StateVariableVisibility,
     StructDefinition,
+    TryCatchClause,
     TupleExpression,
     TypeName,
     UnaryOperation,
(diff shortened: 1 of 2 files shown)


cd1m0 (approved):
lgtm

PR #193, opened 2023-04-05
Add support for integer casts in constant evaluator

cd1m0:
What title says.


blitz-1306 (approved):
A great patch, thanks.

Still we better add an `InferType` instance as an argument due to there is a need to distinct Solidity versions and make logic branches depending on it.

PR #194, opened 2023-04-05
Inject `InferType` instance to `evalConstantExpr()`

blitz-1306:
## Brief
This is a followup PR after #193.

## Changes
- [x] Update dependencies.
- [x] Inject `InferType` instance as an argument for constant expression evaluator. This would allow us to react in behavior difference for various Solidity versions in the future.
- [x] Tweak `InferType.typeOfElementaryTypeNameExpression()` to have more precise return type.
- [x] Fix affected tests.

## Notes
This PR contains API breaking changes. Pending release should increase **major** version of the package to not break the dependent packages.

Regards.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -26,11 +26,11 @@
             "devDependencies": {
                 "@types/fs-extra": "^11.0.1",
                 "@types/mocha": "^10.0.1",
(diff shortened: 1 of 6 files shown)


cd1m0 (approved):
lgtm

PR #198, opened 2023-04-06
Respect types when evaluating unary and binary operations in `evalConstantExpr()`

blitz-1306:
## Changes
- [x] Respect types when evaluating unary and binary operations in `evalConstantExpr()`.
- [x] Add edge-cases to tests.

Regards.

### DIFF
--- a/src/types/infer.ts
+++ b/src/types/infer.ts
@@ -1684,6 +1684,10 @@ export class InferType {
             return innerT;
         }
 
+        if (node.operator === "-" || node.operator === "+" || node.operator === "~") {
+            return innerT;
+        }
+
         if (innerT instanceof NumericLiteralType) {
             const res = evalConstantExpr(node, this);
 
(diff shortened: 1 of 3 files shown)


src/types/infer.ts
```diff
@@ -497,19 +497,6 @@ export class InferType {
         const a = this.typeOf(node.vLeftExpression);
         const b = this.typeOf(node.vRightExpression);
 
-        if (a instanceof NumericLiteralType && b instanceof NumericLiteralType) {
```
cd1m0, line 500: I'm worried about moving this line below. We want `InferType.typeOf(`4+3`)  to be `IntLiteral(7)`. I think moving this below would break this

cd1m0 (changes requested):
I think moving the IntLiteral checks would break type inferrence (see comment below). I think we should come up  with another solution to avoid the inf recursion

src/types/infer.ts
blitz-1306: I restored original logic in d4a7200 and found workaround in 2fcc3d9.

cd1m0 (approved):
LGTM

PR #209, opened 2023-05-30
Function to check that `TypeNode` is ABI-encodable

### LINKED ISSUE #206: Introduce function to check that TypeNode is ABI-encodable
## Description

As part of implementation of https://github.com/ConsenSys/scribble/pull/224, there is a request to introduce function to check if arbitrary expression is ABI-encodable.  The feature would help to do sanity check for source fragments, that can be wrapped by an `abi.encode()` call.

## Context

See [the related behavior](https://github.com/ethereum/solidity/blob/0a0c389541c0f247691951edd64451be7145436c/libsolidity/ast/Types.cpp#L314-L339) is of Solidity compiler. Feature most likely would require adding new method to `InferType`, that accepts `TypeNode`.

blitz-1306:
## Preface
Fixes #206, then we can introduce it to [Scribble's `eq_encoded()`](https://github.com/ConsenSys/scribble/pull/224).

## Changes
- [x] Introduce `InferType.isABIEncodable()` method and related unit test.
- [x] Tweak `InferType.toABIEncodedType()` to check if type is encodable.
- [x] Other small fixes.

Regards.


src/types/infer.ts
```diff
@@ -2341,6 +2343,48 @@ export class InferType {
         return false;
     }
 
+    isABIEncodable(type: TypeNode, encoderVersion: ABIEncoderVersion): boolean {
```
cd1m0, line 2346: If we add some new builtin type, we might forget to add it to this `if` and `isABIEncodable` would automatically assume that that type is encodable. I would prefer if we flip the logic, that we `isABIEncodable` explicitly lists the allowed types, and by default it returns `false` at the end.

cd1m0 (changes requested):
Good work! One nit in the logic

src/types/infer.ts
blitz-1306: Addressed in c4c55d6

cd1m0 (approved):

PR #213, opened 2023-06-08
Improve precision of `isFunctionCallExternal()`

### LINKED ISSUE #171: Increase `isFunctionCallExternal()` precision by using `InferType`
```typescript
export function isFunctionCallExternal(call: FunctionCall, inference: InferType): boolean {
    if (call.kind !== FunctionCallKind.FunctionCall) {
        return false;
    }

    if (
        call.vFunctionCallType === ExternalReferenceType.Builtin &&
        CALL_BUILTINS.includes(call.vFunctionName)
    ) {
        return true;
    }

    const exprT = inference.typeOf(call.vExpression);

    return exprT instanceof FunctionType && exprT.visibility === FunctionVisibility.External;
}
```
Above implementation would be more precise than pending implementation from #159, as it does not rely on typeString lookups.

However, implementing it as in above example is blocked due to failures of unit test. The `exprT` is resolved to a following:
```
FunctionType {
  id: 141,
  src: undefined,
  name: undefined,
  parameters: [],
  returns: [],
  visibility: 'public',
  mutability: 'nonpayable',
  implicitFirstArg: false
}
```
Here is the related Solidity sample:
```solidity
contract Main {
    function c() public {}
    function d() public {}

    function main() public payable {
        (true ? this.c : this.d)();
    }
}
```

This sample is a shortened version of sample in `external_calls.spec.ts` integration test. The reason of failure is that it would need more time and efforts to investigate how to handle such expressions in typing context to figure out that the member access result is an external function (that is somehow done by compiler and reflected in typeString).

Current implementation seem to be fragile in edge-cases if we have struct with word `external` somewhere in the name, if also refered as an argument.

blitz-1306:
## Preface
This PR fixes #171.

## Changes
- [x] Tweak `isFunctionCallExternal()` to rely on type inference instead of analyzing `typeString`s.

Regards.

### DIFF
--- a/src/types/utils.ts
+++ b/src/types/utils.ts
@@ -332,11 +332,26 @@ export function isFunctionCallExternal(call: FunctionCall, inference: InferType)
     const exprT = inference.typeOf(call.vExpression);
 
     if (exprT instanceof FunctionType) {
+        if (exprT.implicitFirstArg) {
+            /**
+             * Calls for using-for are not considered as external.
+             * Currently "implicitFirstArg" is used only for using-for.
+             */
+            return false;
+        }
+
         if (exprT.visibility === FunctionVisibility.External) {
             return true;
         }
 
         if (exprT.visibility === FunctionVisibility.Public) {
+            /**
+             * We have external calls when we have call expression like "expr.fun()",
+             * where "expr" is an identifier, "this" or an expression of ContractType.
+             *
+             * Other expressions, that are invoked via an identifier are causing internal calls,
+             * so we exclude them.
+             */
             return call.vExpression.getChildrenByType(MemberAccess, true).length > 0;
         }
     }
--- a/src/types/utils.ts
+++ b/src/types/utils.ts
@@ -337,7 +337,7 @@ export function isFunctionCallExternal(call: FunctionCall, inference: InferType)
         }
 
         if (exprT.visibility === FunctionVisibility.Public) {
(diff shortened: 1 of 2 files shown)


src/types/utils.ts
```diff
@@ -328,8 +329,16 @@ export function isFunctionCallExternal(call: FunctionCall): boolean {
...
-        return true;
+    const exprT = inference.typeOf(call.vExpression);
+
+    if (exprT instanceof FunctionType) {
+        if (exprT.visibility === FunctionVisibility.External) {
+            return true;
+        }
+
+        if (exprT.visibility === FunctionVisibility.Public) {
+            return call.vExpression.getChildrenByType(MemberAccess, true).length > 0;
```
cd1m0, line 340: Do we handle this case correctly on this line:

1. x.functionName, where x is a variable of some type, T, and we have a `using Lib for T;` in the code

cd1m0 (changes requested):
Just 1 question to consider. Otherwise looks good!

src/types/utils.ts
blitz-1306: This is a good catch. I see that we mark such things with `implicitFirstArg`. It seems to have only use for `using Lib for T` is used. I will try to filter out those and add related test case.

blitz-1306: Addressed in a160ce9

cd1m0 (approved):

PR #215, opened 2023-06-16
Fixes to `isFunctionCallExternal()` after end-to-end test

blitz-1306:
## Preface
Test of pending release against [Scribble](https://github.com/ConsenSys/scribble) uncovered additional edge-cases for external call detection:
- We sometimes may have `ContractType.method()` member accesses for `public` functions, where `ContractType` is one of the inheritance bases. Some of those calls are invoked internally. [Example is here](https://github.com/ConsenSys/scribble/blob/b61c7c15aa58337aa1ce97311496a1c2dbbc02a6/test/multifile_samples/erc20_macro_inheritance/child.sol#L10).
- `FunctionLikeSetType`s are now yet handled, where we have various possible functions via overloading. [Example is here](https://github.com/ConsenSys/scribble/blob/b61c7c15aa58337aa1ce97311496a1c2dbbc02a6/test/samples/callsite_replace_overloading.sol#L13-L15).

This PR makes attempts to address these issues. Related Scribble tuning PR is here: https://github.com/Consensys/scribble/pull/232

## Changes
- [x] Reintroduce `isFunctionCallExternal()` as method of `InferType`.
- [x] Improved detection of calls in external context.
- [x] Added relevant cases to a test sample.

## Notes
This is a breaking change (previous PR #213 was also breaking due to signature change), so pending release would have to bump **major** version of the package to avoid breaking dependencies.

Regards.


src/types/infer.ts
```diff
@@ -2662,4 +2663,90 @@ export class InferType {
...
+                );
+            }
+        }
+
+        if (
+            expr instanceof MemberAccess ||
+            expr instanceof FunctionCallOptions ||
+            expr instanceof FunctionCall
+        ) {
+            return this.isExternalCallContext(expr.vExpression);
```
cd1m0, line 2691: I think this code may break for code like this one:

```
import * as Mod from "mod.sol";

...
   Mod.Library.fun(...);
```

If we keep resolving the `vExpression` of `MemberAccess` eventually we will hit `Mod` instead of `Library`.

src/types/infer.ts
```diff
@@ -2662,4 +2663,90 @@ export class InferType {
...
+
+            if (exprT instanceof UserDefinedType) {
+                if (exprT.definition instanceof ContractDefinition) {
+                    return true;
+                }
+            }
+
+            if (exprT instanceof TypeNameType) {
+                return (
+                    exprT.type instanceof UserDefinedType &&
```
cd1m0, line 2679: I am still confused about this one. I thought there may be cases when `Library.fun(....)` may be internal? Or is it always external if the function is defined as public? Sorry I may be forgetting  here

test/integration/types/external_calls.spec.ts
```diff
@@ -95,14 +114,29 @@ contract Main {
...
         a.add(4);
 
+        payable(0).some();
+        payable(0).other();
+
         Some memory s = Some(Lib.w, fExtPtr);
 
         s.intFn();
         s.extFn();
+
```
cd1m0, line 124: Could you add a test for calling `LibAddr.some(payable(0x0))`?

cd1m0 (changes requested):
One question, some nits. Otherwise looking decent!

test/integration/types/external_calls.spec.ts
blitz-1306: Addressed in 4ea38d2

src/types/infer.ts
blitz-1306: Addressed in 4ea38d2. We consider treating such call as `external`.

src/types/infer.ts
blitz-1306: Addressed in 4ea38d2. See extended if coniditon at the beginning of the function.

cd1m0 (approved):

PR #230, opened 2023-10-19
Handle additional edge cases for `evalConstantExpr()`

blitz-1306:
## Changes
- [x] Support `Buffer` values (for hex literals) in unary math ops.
- [x] Introduce `fixedBytesTypeToIntType()` and `castToType()`. Export utility functions for downstream.
- [x] Tweak `evalUnary()` and `evalBinary()` to cover more edge-cases.
- [x] Add related test cases.

Regards.


test/unit/types/eval_const.spec.ts
```diff
@@ -1320,6 +1320,191 @@ const cases: Array<[string, (factory: ASTNodeFactory) => Expression, boolean, Va
...
+        //             factory.makeFunctionCall(
+        //                 "byte",
+        //                 FunctionCallKind.TypeConversion,
+        //                 factory.makeElementaryTypeNameExpression("type(byte)", "byte"),
+        //                 [factory.makeLiteral("<missing>", LiteralKind.Number, "ee", "0xee")]
+        //             )
+        //         ),
+        //     true,
+        //     true
+        // ],
```
blitz-1306, line 1370: @cd1m0 This is actually interesting and I might need help here.

Currently we apply `clampIntToType()` when type of `BinaryOperation` is `IntType`. However in this case we have `BoolType`, so result is not clamped. Other thing here is that two operands are casted to a common type, that is smaller, than type, that is computed by `InferType.inferCommonType()`, that is relying on `castable()`. In `castable()` we have following entry: https://github.com/Consensys/solc-typed-ast/blob/d4dd707aeb1ecff2627893e26e4c6ae1d6fce286/src/types/utils.ts#L431-L434

But in these cases:
```solidity
bytes2(0xcc00) < byte(0xee);
bytes2(0xff00) == byte(0xff);
```
it seems that result is coerced to a smaller size and this is fine even in latest compilers. I tried following:
```typescript
export function evalBinary(node: BinaryOperation, inference: InferType): Value {
    try {
        const leftT = inference.typeOf(node.vLeftExpression);
        const rightT = inference.typeOf(node.vRightExpression);
        const commonT = inference.inferCommonType(leftT, rightT);

        let left = evalConstantExpr(node.vLeftExpression, inference);
        let right = evalConstantExpr(node.vRightExpression, inference);

        if (commonT instanceof FixedBytesType) {
            left = clampIntToType(toInt(left), fixedBytesTypeToIntType(commonT));
            right = clampIntToType(toInt(right), fixedBytesTypeToIntType(commonT));
        }

        const res = evalBinaryImpl(node.operator, left, right);

        if (leftT instanceof NumericLiteralType && rightT instanceof NumericLiteralType) {
            return res;
        }

        const resT = inference.typeOfBinaryOperation(node);

        if (resT instanceof IntType && typeof res === "bigint") {
            return clampIntToType(res, resT);
        }

        return res;
    } catch (e: unknown) {
        if (e instanceof EvalError && e.expr === undefined) {
            e.expr = node;
        }

        throw e;
    }
}
```
But it solves only one case of two. Would you be able to find some time and take a look?

blitz-1306: Addressed in d22f7d6

cd1m0 (changes requested):
One nit. Will fix

src/types/eval_const.ts
```diff
@@ -396,13 +483,16 @@ export function evalLiteral(node: Literal): Value {
...
 
-        if (subT instanceof IntType && typeof res === "bigint") {
-            return clampIntToType(res, subT);
+        if (subT instanceof NumericLiteralType) {
+            return evalUnaryImpl(node.operator, sub);
         }
 
-        return res;
+        const resT = inference.typeOfUnaryOperation(node);
+        const res = evalUnaryImpl(node.operator, evalConstantExpr(node.vSubExpression, inference));
```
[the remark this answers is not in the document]
cd1m0: Fixed in [6269c0b](https://github.com/Consensys/solc-typed-ast/pull/230/commits/6269c0b303aaf68db07ef6e402c207fc8920a366)

cd1m0 (approved):

PR #236, opened 2023-11-08
Fix type of `OverrideSpecifier.vOverrides` to comply with type of child nodes

blitz-1306:


blitz-1306:
Overall, this fix should be more extensive, as we should tweak `ASTNodeWithChildren` to respect children type `T` more properly (`ownChildren` should be of type `T` instead of `ASTNode`, all relevant methods should be tuned).

cd1m0 (approved):

PR #240, opened 2023-12-10
fix error when typeDescriptions missing

hananbeer:
I'm working on an AST tool and it just happens to be convenient to omit unnecessary fields.
the `typeDescriptions` field isn't really necessary for the AST compilation using `sol-ast-compile` yet when it throws an error when it is missing.

### DIFF
--- a/src/ast/modern/expression_processor.ts
+++ b/src/ast/modern/expression_processor.ts
@@ -10,7 +10,7 @@ export class ModernExpressionProcessor<T extends Expression> extends ModernNodeP
     ): ConstructorParameters<ExpressionConstructor<T>> {
         const [id, src] = super.process(reader, config, raw);
 
-        const typeString: string = raw.typeDescriptions.typeString;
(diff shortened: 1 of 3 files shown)


blitz-1306:
Hello, and sorry for the late response. End of the previous year got me preoccupied. A few questions:

1. Would you be able to provide a use when the issue occurs? The issue is occuring on a certain Solidity sample or compiler version?
2. Can we figure out a test for the changes?

Thank you.

cd1m0:
We try to give as much of the compiler information to downstream consumers as possible. All solc-compiler versions support typeStrings, so we expect and provide them to consumers. It gets more inconvenient for downstream consumers if they always have to check if typestrings are there or not. Their code gets simpler if they can always assume they are there.

Internally, during AST-building we don't depend on typestrings, so you can easily provide an empty string for typeStrings and it should work in your case.

I hope this workaround helps with your use-case!

PR #242, opened 2024-01-08
Switch sources to typed arrays instead of strings

cd1m0:
> [!Warning]
> This is an interface **breaking** change.

Change our implementation to keep source file contents as `Uint8Array`s instead of `string`. This is needed since AST source map [are actually **byte offsets**](https://docs.soliditylang.org/en/v0.8.23/internals/source_mappings.html#source-mappings) into the file. We were erroneously trearting them as string indices, which breaks in the presence of multi-byte Unicode characters.

This involves:

1. Changing a bunch of the `compile*` interfaces to accept `Map<string, Uint8Array>` instead of `Map<string, string>`.
2. Changing the `StructuredDocumentation` pass to correctly compute docs by doing byte indexing.
3. Changing the `ASTWriter` to compute a correct map.
4. Adding tests.

Todo (@blitz-1306 ): I havent looked at inline assembly. Is there anything to do there?


src/compile/utils.ts
```diff
@@ -283,7 +281,7 @@ export async function compileSol(
     const remapping = pathOptions.remapping || [];
     const parsedRemapping = parsePathRemapping(remapping);
 
-    const files = new Map<string, string>();
+    const files: FileMap = new Map<string, Buffer>();
```
blitz-1306, line 284: I think we should not reference `Buffer` here:
```typescript
const files: FileMap = new Map();
```

blitz-1306 (commented):
Great work. Two things to consider:

1. There is a coverage failure.
2. One small nit in the code.

I will try to pick up issues tomorrow.

> I havent looked at inline assembly. Is there anything to do there?

No, I think there is nothing left.

src/compile/utils.ts
blitz-1306: Addressed in 73492b7

blitz-1306 (approved):
Looks ready to merge 73492b7.

blitz-1306 (approved):

blitz-1306 (approved):

PR #265, opened 2024-04-24
Fix for #269

cd1m0:


cd1m0:
I think the failure is intermittent due to tests running in parallel. Cannot reproduce locally

PR #273, opened 2024-09-16
Add support for custom compilers in solc-typed-ast versions

cd1m0:
This pr adds support for custom compiler paths in solc-typed-ast. To use a custom compiler you can specify a verison such as "custom:/<path-to-compiler>":

`sol-ast-compile --compiler-kind native --compiler-version 'custom:/path/to/solc' file.sol`


cd1m0 (commented):
lgtm

PR #278, opened 2025-12-01
Add TypeIdentifiers

d1m0:
This PR adds support for `typeIdentifier` parsing as an alternative to the aging and fragile `InferType`.
Back in the old days, when dinosaurs still roamed the Earth, solc-typed-ast had to handle the legacy AST that was available in 0.4.x series of the Solidity compiler. However given that:

- 5 years have passed since then
- the 0.4.x series since 0.4.13 still support emitting the modern AST
- the code for infering the type of an AST node is large, complex, error-prone and incomplete
- just parsing the type identifiers is trivial

I've decided its time to slowly replace `InferType` with just parsing `typeIdentifier`s. The plan is

1. With version 19.0.0 to add support for typeIdentifiers with the:

- typeIdentifier field on `Expression`, `TypeName` and `VariableDeclaration` nodes
- the `typeOf(node: Expression | TypeName | VariableDeclaration): TypeIdentifier` function

2. In a later version to deprecate `InferType` with a warning, and eventually remove it.

### DIFF
--- a/src/ast/ast_node_factory.ts
+++ b/src/ast/ast_node_factory.ts
@@ -1050,13 +1050,13 @@ export class ASTNodeFactory {
             typeIdentifier = `t_contract$_${target.name}_$${target.id}`;
         } else if (target instanceof StructDefinition) {
             typeString = `type(struct ${target.canonicalName} storage pointer)`;
-            typeIdentifier = `t_contract$_${target.name}_$${target.id}`;
+            typeIdentifier = `t_struct$_${target.name}_$${target.id}`;
         } else if (target instanceof EnumDefinition) {
             typeString = `type(enum ${target.canonicalName})`;
-            typeIdentifier = `t_contract$_${target.name}_$${target.id}`;
+            typeIdentifier = `t_enum$_${target.name}_$${target.id}`;
         } else if (target instanceof UserDefinedValueTypeDefinition) {
             typeString = `type(${target.canonicalName})`;
-            typeIdentifier = `t_contract$_${target.name}_$${target.id}`;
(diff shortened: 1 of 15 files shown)


src/ast/ast_node_factory.ts
```diff
@@ -1022,24 +1044,31 @@ export class ASTNodeFactory {
             }
 
             typeString = result.join(" ");
+            typeIdentifier = undefined;
         } else if (target instanceof ContractDefinition) {
             typeString = `type(contract ${target.name})`;
+            typeIdentifier = `t_contract$_${target.name}_$${target.id}`;
         } else if (target instanceof StructDefinition) {
             typeString = `type(struct ${target.canonicalName} storage pointer)`;
+            typeIdentifier = `t_contract$_${target.name}_$${target.id}`;
```
d1m0: wrong string

src/ast/postprocessing/type_identifier_normalizer.ts
```diff
@@ -0,0 +1,42 @@
...
+            node.parent instanceof ElementaryTypeNameExpression &&
+            typeof node.parent.typeIdentifier === "string"
+        ) {
+            const parentType = parseTypeIdentifier(node.parent.typeIdentifier);
+            if (parentType instanceof TypeTypeId) {
+                node.typeIdentifier = parentType.actualT.pp();
+            }
+        }
+
+        if (node instanceof Identifier && node.parent instanceof ImportDirective && node.typeIdentifier === undefined && node.vReferencedDeclaration instanceof VariableDeclaration) {
```
d1m0: Add comment for this case

